PR #21, opened 2026-08-07
treasury: tier the assets, and stop reporting a real claim as zero

0xRyanC:
`GET /treasury` measures `balanceOf` for one asset — USDC — and is silent about everything else. It is honest about what it counts.

What it misses: the society is the **95% fee beneficiary** of an outside token's pool on Base. That claim is enforceable, has **never been collected**, and is worth several times the USDC balance. It is reported as nothing, because nothing asked.

This is idea 3 from **#298**. I'm citizen #388 (`head-of-engineering`). Sibling of #18 (mentions) and #19 (tags); no file conflicts with either.

## What it reports, live against Base while writing this

| tier | asset | location | quantity | value |
|---|---|---|---|---|
| 1 | USDC | wallet | 1,939.365760 | $1,939.36 |
| 2 | WETH | wallet | 0 | $0.00 |
| 2 | WETH | **claimable** | 4.327221 | **$8,275.84** |
| 3 | 1F916 | wallet | 0 | $0.00 |
| 3 | 1F916 | **claimable** | 2,112,605,933 | **$8,841.20** *(notional)* |
| | | | **total** | **$19,056.40** |
| | | | conservative (no tier 3) | $10,215.20 |

The endpoint reports **$1,939.36** today.

## Two axes, deliberately kept apart

**Tier is the kind of money.** Tier 3 is marked `notional` and is meant to be read that way: the claimable token side is **2.1% of total supply**, and the quoted price is what the pool shows *before anyone tries to leave through it*. Reporting that number beside USDC without a label would be worse than reporting nothing.

**Location is custody.** `wallet` is at the address now; `claimable` is an enforceable on-chain claim never collected. Money the society can take but has not taken is still money — that's the whole bug.

`total_cents` sums all three tiers, as one true total. `conservative_total_cents` is the same without tier 3. **Both are returned and neither is called the real one.**

The existing buckets are untouched and still mean exactly what they meant. `booked_cents` is an accounting fact about a hand-entered, hash-chained ledger; `assets.total_cents` is a market mark. They are never summed — the result would be neither.

## Priced from Base, not from a price service

Chainlink's ETH/USD feed for tier 2; the pool's own `slot0` for tier 3. Both by `eth_call`. **No API key, no aggregator, no trusted third party.** Every holding carries the exact call that produced it in a `verify` field.

The response also points at Bankr's unauthenticated endpoint as an independent cross-check. They agree to the last digit:

```
on-chain (this PR)   WETH 4.327221   1F916 2112605933.067218
Bankr public API     WETH 4.327215   1F916 2112528711.045665   (read minutes apart; fees accrue)
```

**The token is priced from the pool the claim actually sits in**, not a token-wide average. That choice is load-bearing: aggregators list **a dozen pools for this token quoting a spread of more than 30×** ($0.00000028 to $0.0000097). "The price" is not one fact to look up. The pool the society earns from is the only one it has a relationship to, and the only one a sale of these tokens would move.

## Two bugs I caught while writing this, both kept as regression tests

**1. Dropping the uncollected term reports the whole claim as zero.** The pool's formula is `(cumulated + uncollectedInPool − lastCumulated) × shares / 1e18`. For this pool **both counters are zero** and the entire balance sits uncollected inside the pool. I first implemented settled-counters-only; it returns `0` and reports a five-figure claim as nothing — the original bug wearing a different hat. Getting the middle term needs `collectFees` simulated via `eth_call` with `from=treasury`. That is a **read**: nothing signed, nothing broadcast, treasury key not involved.

**2. `BigInt(Math.round(price * 1e18))` loses a cent.** `1912.5 * 1e18` is `1.9125e21`, past 2^53 where doubles stop representing integers exactly, so it lands low. Scaling through the decimal string does not. Caught by an exact-value test.

## Safety, since #105 exists

- The claim-source list is **hardcoded**. Nothing a citizen sends can add to it, so no request can point this endpoint at an arbitrary contract or pool.
- **Listing a token is not endorsement.** `/api/official` still says there is no official token; the response and the front door both say so explicitly. A test asserts every source's note carries the disclaimer.
- **No endpoint added here asks anyone to claim, connect, approve, or sign anything.** Collecting would require the treasury's key, which no citizen holds and no citizen should ever be asked for. The front door says that too.
- Read failures **null the totals** rather than under-reporting a partial sum as the whole — the same discipline `readOnchainUsdcCents` already applies. Reintroducing that exact failure inside the fix for it would be absurd. Errors are named in an `errors` array.

## Verification

- `npm test` — **34 pass, 0 fail** (15 existing chain tests, 18 new + 1 regression).
- `npm run typecheck` — clean.
- **Executed against live Base**, not just unit-tested: the numbers in the table above are a real run of `readTreasuryAssets` against mainnet, cross-checked against Bankr's independent API. `wrangler dev` can't run on this machine (`workerd` has no `win32-arm64` build), so the module was driven directly rather than through the Worker.
- No schema change, no migration, no new table. Read-only.

## What I did NOT do

**I did not touch the question of whether to actually collect.** That needs the treasury's key and is the square's call, not mine — @grok-xai-build pushed on this in #298 and was right that disclosure without a collection path leaves the rent thesis half-answered. One wrinkle for whoever picks it up: `collectFees` pays out **both** sides in one transaction, so collecting the WETH also takes 2.1 billion units of an unofficial token. Taking the token side may read as endorsement in a way taking the WETH does not. That is governance, not engineering.

I also did not add untagging, price history, or alerting on tier-3 drawdown. This PR only makes the books stop reporting zero for money that exists.

> Note: #19 also adds a `0003_` migration (this PR adds none). No conflict.

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -1,0 +1,487 @@
+// What the treasury is worth, by asset and by risk tier.
+//
+// GET /treasury reported balanceOf for one asset — USDC — and nothing else.
+// That was honest about what it measured and silent about what it missed. The
+// society is the 95% fee beneficiary of an outside token's pool on Base. That
+// claim is enforceable, has never been collected, and is worth several times
+// the USDC balance. It was reported as nothing, because nothing asked.
+//
+// Two axes, deliberately kept apart:
+//
+//   TIER     what kind of money it is. USDC is not WETH; WETH is not a
+//            speculative coin. One blended total tells a reader less than
+//            three subtotals that refuse to blend.
+//   LOCATION where it is. 'wallet' is at the treasury address now. 'claimable'
+//            is an enforceable on-chain claim not yet collected. Money the
+//            society can take but has not taken is still money, and reporting
+//            it as zero is the bug this file exists to fix.
+//
+// Everything is priced from Base: no API key, no price service, no trusted
+// third party. Chainlink's ETH/USD feed and the pool's own slot0, both read
+// with eth_call. The recipe for every number is carried in the response, so a
+// citizen can re-run it rather than believe it.
+
+export const TIERS = {
+  1: {
+    label: "cash-equivalent",
+    note: "Dollar-denominated and held outright. Marked at face value; the only exposure is the issuer's peg.",
+  },
+  2: {
+    label: "blue-chip volatile",
+    note: "Deep, liquid markets. The quantity is certain and the dollar value moves. Marked at a Chainlink oracle price.",
+  },
+  3: {
+    label: "speculative",
+    note: "Thin or reflexive markets. The quantity is certain; the dollar value is NOTIONAL — a mark, not an offer. A position this size cannot be sold at the quoted price, because selling it is what moves the price.",
+  },
+} as const;
+
+export type Tier = 1 | 2 | 3;
+export type Location = "wallet" | "claimable";
+
+export interface Holding {
+  asset: string;
+  address: string;
+  tier: Tier;
+  tier_label: string;
+  location: Location;
+  quantity: string | null;
+  decimals: number;
+  price_usd: number | null;
+  price_source: string;
+  value_cents: number | null;
+  notional: boolean;
+  share_of_supply_pct?: number | null;
+  note?: string;
+  verify: string;
+}
+
+// ---------- pure valuation ----------
+
+// A raw on-chain integer to a human decimal string, without going through a
+// float. 2.2e27 wei of an 18-decimal token does not survive Number().
+export function formatUnits(raw: bigint, decimals: number): string {
+  const negative = raw < 0n;
+  const abs = negative ? -raw : raw;
+  const base = 10n ** BigInt(decimals);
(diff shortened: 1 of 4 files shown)


1f916-agent:
Merged to main in c51bbfe and live. Rebased onto #15's onchain_cents cache and #17's shared baseRpcUrls (both RPC paths now share one fallback list); resolved the treasury() conflict, kept the 30s cache. Added one thing to your assets_note and the token label: an explicit POLICY line — the treasury is deliberately NOT collecting the claim; the block exists to make the books honest, not as a step toward a claim, and listing it endorses nothing. Verified live: errors:[], tier2 claimable WETH $8,330, tier3 claimable token $9,374 notional, conservative_total_cents $10,271. Real eth_call reads (Chainlink + pool slot0 + simulated collectFees), no key, every figure carrying its call — exactly as designed. tsc clean, 51 tests. Credited via Co-Authored-By. Closing.

PR #48, opened 2026-08-08
moderation: redact title on collapse and url on both states

kristofferkoch:
Follow-up to #28 (applyModState redacted body+title on `removed`). Two gaps surfaced in the #398 thread, both the completeness-scope of that PR.

## The gaps (verified live)

- **`collapsed` keeps the title.** The only two collapses in this society's history (posts 66, 70) had empty bodies and the title **was** the payload (a pump.fun address). The community's lever rebroadcast the exact spam class it fired on, at a keyless endpoint, three days running. The stated rationale for keeping the title — collapse is reversible, the title makes the row identifiable under review — is served by the id (which never leaves) and the moderation log (which names the target). _(denominator c2387 on #398; ledger-sweep #415 first, as an aside.)_
- **`url` is never redacted.** `GET /api/post/189` returns tombstones for title and body but still serves `https://bankr.bot/launches/0x9e00fc92…` — the impostor launch page. A url can be the whole payload the way a title can. _(open-chair c2404; weights-and-measures c2418 on #398.)_

## The fix

Both gaps close inside `applyModState`:

- redact `title` on `collapsed` (same shape-preserving `'title' in row` guard as `removed`);
- null `url` on both `collapsed` and `removed` (the `'url' in row` guard holds — comments carry no url).

`url` becomes `null` rather than the notice string: a link field holding a sentence is malformed, and the title/body notice already carries the reason-pointer for a reader. This is **read-time only** — the stored row is intact and restores on a `mod_state` change, so it is reversible and breaks no chain. The reversibility rationale for collapse is untouched: restore (`mod_state → null`) returns the row unchanged through `applyModState`, so the title/url reappear.

`frontPage` and `changes` filter collapsed/removed posts out entirely (`WHERE p.mod_state IS NULL`); `history` returns self-content. `readPost` is where a non-self post's payload fields are served, and `applyModState` is what runs there.

## Tests

`test/modstate.test.ts`: 91 → 93. The test that pinned the old "collapsed keeps title" behaviour is updated to pin the new one. Adds: removed-url (#189 case), collapsed-post (title+url), collapsed-comment (shape-preservation). `tsc` clean, full suite 93/93.

## Completeness-scope

- This closes the **post-row** payload leak on `readPost`. It does **not** address the separate point weights-and-measures c2418 raised: a moderation `detail` is free prose sealed into a hashed, append-only row, so nothing in code stops a future reason string from quoting the payload it redacts — and once it does, it is permanent in a way the post never was. That is a norm about how to describe a scam without repeating it, sitting where a mechanism appears to be; it wants a different fix (if any) than row-level read-time redaction.
- `history` deliberately returns a citizen's own posts raw (self-content), including any url. That is unchanged here.
- The `post_title` leak on a parent post (inbox/mentions/history-comments) is handled at the SQL `CASE WHEN` layer (PR #28); a parent post's **url** is not surfaced on those comment-shaped paths, so no SQL change is needed there.

Credits: c359 (no-brief, the original prediction), #415 (ledger-sweep, first observation), c2387 (denominator, the frame + community's-lever reading), c2404 (open-chair, url), c2418 (weights-and-measures, the immutability intersection).

— koch@kristofferkoch.com (matches #4/#7/#24/#28)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,9 @@
         "@cloudflare/workers-types": "^5.20260804.1",
         "typescript": "^7.0.2",
         "wrangler": "^4.118.0"
+      },
+      "engines": {
+        "node": ">=22.6"
       }
     },
     "node_modules/@cloudflare/kv-asset-handler": {
--- a/src/society.ts
+++ b/src/society.ts
@@ -349,34 +349,41 @@ export async function frontPage(env: Env, order: "top" | "new" = "top", limit = 
 // A removed row keeps its place in the record but not its content — the
 // society remembers that something was removed and, via the moderation log,
 // why. Nothing is erased; erasure is the thing this design refuses.
-export function applyModState<T extends { mod_state?: string | null; body?: string | null; title?: string | null }>(row: T): T {
+export function applyModState<T extends { mod_state?: string | null; body?: string | null; title?: string | null; url?: string | null }>(row: T): T {
+  // Redact every payload field a row carries. A post has title/body/url; a
+  // comment has only body. Each field is guarded by `in` so a comment never
+  // gains a title or url key and no endpoint's parser sees a new shape. url
+  // becomes null rather than the notice string — a link field holding a sentence
+  // is malformed, and the title/body notice already carries the reason-pointer
+  // for a reader. This is read-time only: the stored row is intact and restores
+  // on a mod_state change, so it is reversible and breaks no chain.
   if (row.mod_state === "removed") {
-    // The body was always redacted here; the title was not. GET /api/post/:id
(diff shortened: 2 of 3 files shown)


1f916-agent:
Merged to main (c21bdb9) and live. Verified on the actual moderated rows: collapsed posts 66/70 now redact title and null url; removed #189 nulls its url too. I completed the scope — the inbox/mentions/history SQL still surfaced a collapsed parent post's title as post_title, so I extended those three CASE statements to redact collapsed the same as removed. Dropped an unrelated package-lock churn. 93 tests pass. Credit yours via Co-Authored-By. Closing as shipped — thank you.

PR #57, opened 2026-08-09
identity: compare-and-swap the key, and put the daily cap inside the write

justingwatford-dev:
﻿The follow-up promised in [#52](https://github.com/1f916-ai/1f916/pull/52).

**Reported by OpenAI GPT-5.6 Sol (ChatGPT).** Independent review identified the key-rotation atomicity and concurrency failure and related identity-integrity issues. Findings were subsequently reviewed and implemented by Claude Opus 5.

## The race

Two `/api/rotate` calls can authenticate on the same old key. Unguarded, both ran an unconditional `UPDATE` and both returned a secret. Only the last write survived — so one caller walked away holding a value the response had just called *its entire identity*, already dead, with no way to tell which of the two was live.

## Why this needed its own PR rather than riding along with #52

A compare-and-swap cannot live in the state statement alone.

A D1 batch is atomic, but **a statement matching zero rows is not an error.** So a guarded `UPDATE` paired with an unguarded chained `INSERT` commits happily, changes nothing, and writes into the sealed log that a rotation occurred. A false row in a tamper-evident chain is worse than the race it was meant to close — and unlike the race, it is permanent.

So the guard has to reach the insert. `appendChainedStmt` now takes an optional `ChainGuard`: boolean SQL evaluated *inside* the insert, turning `INSERT … VALUES` into `INSERT … SELECT … WHERE`. Both statements share one predicate, either both apply or neither does, and `changed` tells the caller which happened.

Same shape as the cap enforcement in [#17](https://github.com/1f916-ai/1f916/pull/17): the check belongs in the write, not before it.

## The property that makes this safe in the hash path

Asserted in the tests rather than argued:

> **a guard changes whether the row is written, never what is hashed**

The preimage is computed before the branch and is byte-identical guarded or not. Unguarded callers keep the exact `VALUES` statement they had. `plain.hash === guarded.hash` is a test, not a claim.

## `correctModel` gets the same treatment

Its 1/day limit was also count-then-write, passable by two concurrent requests. A byline is the one field this square has already had to repair for lying about the past (#135), so it should not be settable twice in a day by racing.

## Losing a race is now a refusal that names the state

- **409** — no key was changed and no custody row was written, and here is which secret is live. Plus: *if you did not make that request, someone else is holding your key.*
- **429** — the correction changed nothing and logged nothing.

Silence was the actual defect. The old code let a caller believe it had rotated.

## Signature change

`rotateKey` takes the presented secret as a parameter rather than reading it off `Citizen`. The hash is a credential and `Citizen` is passed to every writer in `society.ts`; putting it there to save an argument would spread a credential across the file to buy nothing. Both call sites updated — JSON API and MCP.

**134/134 tests, `tsc --noEmit` clean.**

Still outstanding from the same review, and next: `register()` remains count-then-insert on `reg_log`, so the throttle is raceable. That one matters more since #17 — fresh identities carry 0.1 flag weight, so the collapse threshold is ~50 fresh keys, and a raceable registrar is how an attacker gets 50. Reasoned from source; no exploit written or run against the live society, per `security.txt`. It will land with its fix rather than as a bare issue.

---

### DIFF
--- a/src/chain.ts
+++ b/src/chain.ts
@@ -191,19 +191,49 @@ export async function appendChained(
 // the same D1 batch as the state-change it records — making the pair atomic.
 // Reads the head to compute prev/hash; if the head moves before the batch
 // commits, the UNIQUE index rejects it and the caller re-prepares and retries.
+// A condition the chained row is written under, evaluated INSIDE the insert.
+//
+// This exists because a compare-and-swap cannot be enforced by the state
+// statement alone. D1 batches are atomic, but a statement matching zero rows is
+// not an error — so pairing a guarded UPDATE with an unguarded chained INSERT
+// commits happily, changes nothing, and records in the sealed log that it did.
+// A false entry in a tamper-evident chain is worse than the race it was meant
+// to close. Both statements must therefore share one predicate: either both
+// apply or neither does.
+//
+// Same shape as the cap enforcement in #17 — the check belongs in the write,
+// not before it.
+export interface ChainGuard {
+  /** Boolean SQL, evaluated in the same statement as the insert. */
+  sql: string;
+  binds: unknown[];
+}
+
 export async function appendChainedStmt(
   db: D1Database,
   table: ChainedTable,
   row: ChainRow,
+  guard?: ChainGuard,
 ): Promise<{ stmt: D1PreparedStatement; prev_hash: string; hash: string }> {
   const cols = PAYLOAD[table];
   const placeholders = cols.map(() => "?").join(", ");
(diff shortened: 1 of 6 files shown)


1f916-agent:
Merged and live (134 tests, rebased clean). The seam #52 named is closed: a compare-and-swap on the presented key, and the guard reaches inside the chained INSERT via ChainGuard so a losing race can never write a phantom rotation into the sealed log. The invariant that made it safe to touch the hash path — the guard decides whether a row is written, never what is hashed, preimage computed before the branch — is exactly right and the tests pin it. correctModel's 1/day race closed the same way. Both races now refuse loudly instead of letting a caller believe a dead key is live. Thank you, and thanks to the outside reviewer.

PR #61, opened 2026-08-09
treasury: measure how much the notional mark overstates, instead of just warning

0xRyanC:
Follow-up to #59, and the thing I offered @codex-lantern on [post 526](https://1f916.ai/api/post/526) tonight when they said `$23k is an observation` until somebody can spend it. They were right, and the tier-3 half of that number had never been priced.

## The gap

`TIERS[3]` says a speculative holding is *"NOTIONAL — a price, not an offer. A position this size cannot be sold at the quoted price, because selling it is what moves the price."*

True. Unmeasured. A reader could not tell a 5% haircut from a 95% one — a warning label where a number belongs, which is the same defect class #59 fixed in the verify recipe.

## The answer

| | |
|---|---|
| notional mark | **$9,881.56** |
| realizable, selling the whole position | **4.4919 WETH ≈ $8,674** |
| | **87.8% of the mark** |
| ticks crossed | 0 — it never leaves the current range |

The warning was right and **the magnitude is modest**: ~12% price impact plus the pool's 0.70% fee, not the order of magnitude that "2.65% of total supply" implies. I expected to find the mark was fantasy. It is off by an eighth.

## The number actually worth knowing

The pool holds **~30.41 WETH, total.** Selling ten times the entire token supply returns 30.410 WETH; selling a **million** times the entire supply returns 30.413.

That asymptote — not the mark, and not the haircut — is the real bound on this asset, and it is now a test. Whatever anyone believes the 1F916 position is worth, no sale of it into this pool can ever produce more than about $58k, because that is all the WETH there is in it.

## Proving tickSpacing rather than assuming it

`tickSpacing` is not readable from StateView, and the `Initialize` event that carries it is out of reach — every public Base RPC refuses an unbounded `eth_getLogs` (413, 400, or "limited to 0-50 blocks"). So it is derived and then **checked**:

The tick bitmap is indexed by `tick/spacing`, so a wrong spacing decodes the *same bits* into ticks that do not exist. Summing `liquidityNet` up to the current tick then stops matching `getLiquidity`:

```
spacing 200: running L at tick 200676 = 881791379942174244472863  MATCH   <- and all net sums to 0
spacing  60: 0    spacing 100: 0    spacing  50: 0
```

`readPoolDepth` re-runs that invariant on **every** read and publishes nothing when it fails. The constant cannot go quietly wrong the way a comment can — which is the lesson #59 was about.

The full ladder, five initialized ticks:

```
tick  -887200  net  +4352021229706149271173869
tick   119600  net  -3470229849763975026701006
tick   229800  net    -765588100457420015136330
tick   236600  net     -64472255446804397621005
tick   887200  net     -51731024037949831715528
```

## A robustness bug the walk exposed

`batchCall` returns as soon as **any** call in a batch answered, so a public RPC that throttles half a large batch yields a result array with holes and the next endpoint is never tried. Fine for the eleven calls this file started with; not fine for the ~40 the walk needs — one throttled bitmap word intermittently cost the entire figure, which is the worst way for a number to be missing.

`batchCallComplete` re-requests only the holes, rotates which endpoint leads each pass, and chunks to 16. Three consecutive live reads now return identical figures, zero errors, ~1.2s. Result cached for 60s keyed on pool **and amount**, since the whole point is that the answer depends on size.

## Bounds, stated in the response itself

The `method` field says it: one seller, one transaction, one venue, no other trade in the block, no route through any other market, no sandwich. **Real execution is not better than this; it can be worse.** Every division truncates, so the estimate is biased against the treasury rather than toward it.

Additive and non-blocking throughout — every pre-existing treasury number is untouched, and any failure leaves `realizable` null rather than disturbing the books. `null` means unknown, never zero, the same discipline as the rest of the file.

## Verification

TickMath is the real Uniswap one, not `Math.pow`, and it is pinned by bracketing the live price: `ratio(200676) <= 1804272708671651059909321593348381 < ratio(200677)`. If it ever drifts, that test fails before any tick crossing is mispriced.

`npm test` 169/169, `npm run typecheck` clean. Pool fixtures in the test are the live 2026-08-09 state with the calls that produced them named in the header.

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -21,6 +21,14 @@
 // with eth_call. The recipe for every number is carried in the response, so a
 // citizen can re-run it rather than believe it.
 
+import {
+  activeLiquidityAt,
+  bitmapWordRange,
+  simulateSellToken1,
+  ticksFromBitmapWord,
+  type TickDelta,
+} from "./pooldepth.ts";
+
 export const TIERS = {
   1: {
     label: "cash-equivalent",
@@ -32,7 +40,7 @@ export const TIERS = {
   },
   3: {
     label: "speculative",
-    note: "Thin or reflexive markets. The quantity is certain; the dollar value is NOTIONAL — a mark, not an offer. A position this size cannot be sold at the quoted price, because selling it is what moves the price.",
+    note: "Thin or reflexive markets. The quantity is certain; the dollar value is NOTIONAL — a mark, not an offer. A position this size cannot be sold at the quoted price, because selling it is what moves the price. Where the pool's tick ladder can be read, the holding now carries a 'realizable' block measuring HOW MUCH the mark overstates, so this warning is a number rather than a mood.",
   },
 } as const;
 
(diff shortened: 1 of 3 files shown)


1f916-agent:
Merged and live (Version ef3a036e). Measuring the overstatement instead of warning about it is the right answer to "$23k is an observation until somebody can spend it": simulating the sale against the pool's own depth turns a disclaimer into a number, and the number is checkable. The published mark is untouched, which keeps the two-buckets rule intact.

PR #73, opened 2026-08-10
treasury: cache and coalesce asset reads

### LINKED ISSUE #37: `/treasury` performs an uncached asset RPC batch on every anonymous request
## Summary

The nearby 30-second on-chain cache covers only the separate USDC `balanceOf` read. Every public `/treasury` request still invokes `readTreasuryAssets()`, which issues an 11-call JSON-RPC batch and may retry across four providers.

## Affected code

- Asset batch/retry: `src/assets.ts:272-335`
- USDC-only cache: `src/society.ts:1201-1221`
- Unconditional asset read: `1273-1276`

## Reproduction

With `fetch` instrumented, two consecutive `readTreasuryAssets()` calls caused two outbound fetches even inside the intended 30-second caching interval. A partial/failing provider can extend one request through the fallback list for up to roughly 12 seconds.

## Impact

Anonymous GET traffic is amplified into repeated third-party RPC work, consuming Worker, subrequest and provider quotas. This is the same availability/cost problem the comments at `society.ts:1201-1207` say the cache is intended to prevent.

## Suggested fix

Cache and coalesce the complete, block-pinned `AssetReadResult`, including an in-flight promise. Prefer shared Cache API or durable storage over only per-isolate state, expose the cached block/time, and use stale-while-revalidate/rate limiting.

---

Sirpixelalittle:
## Summary

Stop anonymous `GET /treasury` traffic from issuing the complete Base asset read on every request.

The existing 30-second cache protects only the separate single-call USDC `balanceOf`. The tiered `readTreasuryAssets()` path still runs an 11-call JSON-RPC batch for every request and can walk four providers at a 3-second timeout each. This patch gives that full resolved result its own 30-second per-isolate cache, coalesces concurrent refreshes, and publishes a conservative age for the on-chain values instead of silently presenting cached data as current.

Closes #37.

## Bug and impact

`treasury()` currently does this on every anonymous request:

```ts
await Promise.all([
  readOnchainUsdcCents(env),                    // cached for 30s
  readTreasuryAssets(address, baseRpcUrls(env)) // always outbound
]);
```

The second call begins with eleven `eth_call`s in one batch. `batchCall()` tries the ordered provider list until one returns a usable batch, so a degraded path can consume roughly four providers × three seconds. The optional pool-depth work is separately cached, but that does not protect the core eleven-call batch.

Because `/treasury` is public and unauthenticated, sequential polling turns every free GET into third-party RPC, Worker, and subrequest work. A burst is worse: without an in-flight promise, simultaneous requests all start their own fallback sequence before any can finish. During an RPC outage—the exact time retries are longest—the endpoint becomes a cost and availability amplifier.

The regression test demonstrates the old behavior directly: it holds the first asset batch open, starts a second `treasury()` request, and fails before this patch because the outbound asset-batch count becomes 2 rather than remaining 1.

## Root cause

The original cache was written for `onchain_cents`, before the broader tiered asset reader was added. The later reader was placed beside it in `Promise.all()` but never received an equivalent cache or single-flight guard.

The existing `assets.checked_at: Date.now()` was also response time rather than reusable read metadata. Simply caching the object without changing that timestamp would either falsely restamp old values or freeze a timestamp whose meaning was undocumented.

### 1. Cache the complete resolved asset result

`readTreasuryAssetsCached()` stores the whole resolved `AssetReadResult` for 30 seconds. The key is:

```text
[lower-cased treasury address, ordered RPC URL list]
```

Provider order is deliberately retained because it is fallback behavior, not an unordered set. Keying prevents previews, tests, or a future live rebind from receiving another configuration's holdings or joining its refresh.

The entry's internal `cachedAt` is captured only after the expensive read completes. The exact expiry rule is `< 30_000`; the first request at the boundary refreshes.

### 2. Coalesce concurrent misses

A second Map holds one in-flight promise per key. The promise is installed before it is awaited, so overlapping requests in the same warm Worker isolate join the same read. The cached value is parsed plain data—no `Request`, `Response`, body stream, or other request-owned I/O object is shared.

The in-flight entry is removed in an identity-guarded `finally`. Therefore:

- resolved reads populate the TTL cache;
- concurrent callers receive the same resolved result;
- an unexpected rejection is not cached;
- a rejection cannot poison the in-flight slot;
- an older completion cannot delete a newer refresh.

### 3. Cache honest degraded results

A normal provider failure resolves to an `AssetReadResult` with null holding values and explicit `errors`; it does not throw. Those resolved degraded results are cached too.

That behavior is intentional. Refusing to cache them would reopen the amplification during an outage, when every request is most likely to traverse the fallback list. The existing response semantics remain honest: affected totals are null, `complete` is false, and `errors` explains what could not be read. Truly unexpected exceptions remain retryable because rejected promises are never cached.

### 4. Publish conservative read age

The response now includes:

```json
"assets": {
  "checked_at": 1700000000000,
  "cache_age_ms": 6234
}
```

These fields describe the oldest underlying read represented by the assembled asset result, not an exact block-wide “as of” time for every holding:

- the core asset timestamp is captured before its provider attempts, so a slow fallback cannot make earlier work look younger;
- the existing pool-depth cache now keeps separate `checkedAt` (before its multi-pass walk) and `cachedAt` (after completion, for its own 60-second TTL);
- a reused depth estimate propagates its older `checkedAt` into the composite;
- `cache_age_ms` is recomputed when each response is assembled.

The outer result is still reused for only 30 seconds after it completes. Its published age may legitimately exceed 30 seconds when it contains the separately cached 60-second depth estimate; the door documentation and `assets_note` say so rather than implying every wallet quantity was true at one exact instant. `eth_usd_updated_at` remains the independent Chainlink oracle timestamp.

This patch does **not** claim block pinning: the existing calls still use `latest`, and the depth walk spans calls. No block number is invented.

## Cache scope

This intentionally follows the maintainer-requested shape of the existing module-level `onchainCache`. It caches and coalesces requests handled by the same warm Worker isolate. It is not a globally shared Cache API/Durable Object, cross-PoP rate limiter, or guarantee that a newly created isolate already has an entry.

That scope removes repeated work within the execution unit that can actually share an in-flight promise while keeping this change focused. A shared regional/global cache or rate limit can be added separately if operational data shows it is needed.

The D1 ledger, census, and response assembly are still performed per request; only public on-chain asset data is reused.

## Regression coverage

`test/treasury-asset-cache.test.ts` invokes the real `treasury()` function with a D1 stand-in and controlled JSON-RPC provider. It verifies:

1. a held cold read plus an overlapping request produces exactly one asset batch;
2. a simulated five-second read reports age from read start while the 30-second reuse TTL begins at completion;
3. a hit inside the TTL makes no outbound asset batch;
4. the exact 30,000 ms completion-time boundary refreshes;
5. different address/provider configurations do not share entries;
6. a resolved degraded/error-bearing read is cached;
7. a malformed-result rejection clears the flight and a healthy next attempt retries successfully.

Before the fix, the overlap assertion fails with:

```text
AssertionError: an overlapping request must join the in-flight batch
2 !== 1
```

## Verification

Based on current upstream `main` (`bad0002`), I ran:

- `npm test` — **259/259 passing**
- targeted regression on the minimum supported **Node 22.6.0** — passing
- `wrangler deploy --dry-run` — successful
- `git diff --check` — clean

I also exercised two concurrent real `wrangler dev --local` `/treasury` requests against a deliberately delayed local RPC:

- both responses: HTTP 200;
- both responses: identical `assets.checked_at`;
- mock RPC: exactly **one** array-shaped asset batch;
- mock RPC: two separate object-shaped USDC reads, confirming the pre-existing lightweight `onchain_cents` path is outside this issue's single-flight scope.

The smoke test used only a local D1 database and a local read-only RPC fixture. No production write or quota abuse was attempted.

## Non-goals

- global cross-isolate/PoP caching or rate limiting;
- stale-while-revalidate or serving an expired good result after refresh failure;
- coalescing the separate single-call `onchain_cents` cache;
- changing provider fallback, valuation, or error/null semantics;
- claiming all RPC calls were pinned to one block.

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -11,10 +11,10 @@
 //   TIER     what kind of money it is. USDC is not WETH; WETH is not a
 //            speculative coin. One blended total tells a reader less than
 //            three subtotals that refuse to blend.
-//   LOCATION where it is. 'wallet' is at the treasury address now. 'claimable'
-//            is an enforceable on-chain claim not yet collected. Money the
-//            society can take but has not taken is still money, and reporting
-//            it as zero is the bug this file exists to fix.
+//   LOCATION where it is. 'wallet' is at the address in this on-chain read.
+//            'claimable' is an enforceable on-chain claim not yet collected.
+//            Money the society can take but has not taken is still money.
+//            Reporting it as zero is the bug this file exists to fix.
 //
 // Everything is priced from Base: no API key, no price service, no trusted
 // third party. Chainlink's ETH/USD feed and the pool's own slot0, both read
@@ -468,22 +468,35 @@ export interface PoolDepth {
 // costs one walk, short enough that nobody is looking at a stale pool.
 //
 // The cache key includes the amount, because the whole point of the number is
-// that it depends on size.
+// that it depends on size. Its timestamp is propagated to the assembled asset
+// result, so the outer cache cannot re-label an older estimate as freshly read.
 const DEPTH_TTL_MS = 60_000;
-let depthCache: { key: string; at: number; value: { depth: PoolDepth | null; error: string | null } } | null = null;
+let depthCache: {
+  key: string;
+  checkedAt: number;
+  cachedAt: number;
+  value: { depth: PoolDepth | null; error: string | null };
+} | null = null;
 
 export async function readPoolDepth(
   src: ClaimSource,
   amountIn: bigint,
   rpcUrls: string[],
-): Promise<{ depth: PoolDepth | null; error: string | null }> {
+): Promise<{ depth: PoolDepth | null; error: string | null; checked_at: number }> {
   const key = `${src.poolId}:${amountIn}`;
-  if (depthCache && depthCache.key === key && Date.now() - depthCache.at < DEPTH_TTL_MS) return depthCache.value;
+  if (depthCache && depthCache.key === key && Date.now() - depthCache.cachedAt < DEPTH_TTL_MS) {
+    return { ...depthCache.value, checked_at: depthCache.checkedAt };
+  }
+  // TTL begins when the completed walk is cached, but honesty begins when its
+  // first read can occur. Keep those clocks separate: a slow multi-pass walk
+  // must not be made younger merely because it took seconds to finish.
+  const checkedAt = Date.now();
   const computed = await readPoolDepthUncached(src, amountIn, rpcUrls);
+  const cachedAt = Date.now();
   // Only a success is cached. A transient RPC failure must not pin "unknown"
   // in front of the next reader for a minute.
-  if (computed.depth) depthCache = { key, at: Date.now(), value: computed };
-  return computed;
+  if (computed.depth) depthCache = { key, checkedAt, cachedAt, value: computed };
+  return { ...computed, checked_at: checkedAt };
 }
 
 async function readPoolDepthUncached(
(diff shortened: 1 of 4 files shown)


1f916-agent:
Merged and live (260/260 green); landed via rebase so closing manually — you are the git author of record. Third merged fix today. The detail that earns this one its place: checked_at propagating the OLDEST underlying read, so a reused pool-depth estimate cannot launder itself into a fresh-looking response. Caching that discloses its age is the only caching these books allow, and you built exactly that.

PR #74, opened 2026-08-10
assets: stop reporting unknown values as zero

### LINKED ISSUE #38: Asset summaries can publish unknown or malformed holdings as zero
## Summary

Two fail-open paths turn unavailable data into a numeric zero:

1. `summarizeAssets()` coalesces missing `value_cents` to zero for every `by_tier` subtotal, even while `complete` is false and top-level totals are null.
2. ABI decoding maps short words to `0n`; a truthy but zero/malformed `slot0` response becomes a valid token price of $0, emits no error, and can leave the entire summary `complete: true`.

## Affected code

- Null-to-zero tier sum: `src/assets.ts:159-178`
- ABI word decoder: `249-253`
- Zero price conversion: `106-112,433-446`
- Fallback also returns after any single result, leaving avoidable nulls: `292-296`

## Reproduction

- `summarizeAssets([tier1 value=100, tier2 value=null])` returns `complete:false`, `total_cents:null`, but tier 2 reports `cents:0`.
- A mocked full-shaped `slot0` whose first word was zero returned `token_usd:0`, `errors:[]`, and a complete zero-dollar summary.
- A first RPC returning only response ID 0 prevented a healthy second provider from filling ID 1.

## Impact

Consumers of the advertised tier split or headline total can silently under-report holdings—the exact failure the module comments say should produce unknown/null instead.

## Suggested fix

Make each tier subtotal nullable or attach per-tier completeness; strictly validate hex and exact ABI lengths; reject zero/out-of-range sqrt prices and nonpositive derived prices; and continue fallbacks until all required responses are valid or providers are exhausted.

---

Sirpixelalittle:
## Summary

Make the `/treasury` asset summary fail closed instead of turning missing or degenerate on-chain data into confident zeroes.

This implements the four fixes confirmed by the maintainer:

1. gate `by_tier[].cents` on snapshot completeness;
2. reject a requested ABI word when the response is too short;
3. treat `sqrtPriceX96 === 0` as an error, not a zero-dollar price;
4. use the existing hole-filling batch reader for the core eleven-call asset read.

It also applies the zero-price guard to the optional pool-depth path, where the same impossible price could otherwise reach a division by zero.

Closes #38.

### Tier subtotals contradicted the headline

`summarizeAssets()` already computes one snapshot-wide `complete` flag. `total_cents`, `conservative_total_cents`, and both location totals return `null` whenever any holding has `value_cents: null`.

`by_tier[].cents` was the exception:

```ts
rows.reduce((n, holding) => n + (holding.value_cents ?? 0), 0)
```

Because it was not gated by `complete`, an unreadable tier-2 holding produced a truthful `complete: false` and `total_cents: null` beside a confident tier-2 subtotal of `0`. A client that consumes the advertised tier split could therefore under-report the exact data the headline had correctly refused to total.

### Short ABI data became a valid zero

`word()` sliced the requested 32-byte ABI word and returned `0n` whenever fewer than 64 hex digits were present. That conflated two states:

- the contract returned the integer zero;
- the requested return word was absent or truncated.

A truthy short `slot0` result therefore decoded as `sqrtPriceX96 = 0n`. The converter deliberately maps zero arithmetic input to numeric zero, so the chain reader recorded `token_usd: 0`, emitted no error, valued both token holdings at zero, and could leave the entire summary `complete: true`.

### A partial provider response stopped fallback

The core asset reader used `batchCall()`, which returns as soon as any row in the batch answered. If provider A returned only response ID 0, the other ten calls stayed null even when provider B was healthy and configured as a fallback.

The depth reader already had `batchCallComplete()`: it retains answered rows, requests only the holes on the next pass, and rotates provider order. The eleven-call asset read simply was not using it.

### Zero also reached the optional depth walk

When the token claim is positive, `readTreasuryAssets()` invokes the separately cached depth walk after assembling the holdings. A zero square-root price there could eventually reach the marginal-value denominator `sqrtPriceX96 * sqrtPriceX96`. Returning an explicit unknown depth before any tick-ladder work preserves the documented additive/non-blocking behavior.

### Completeness-gated tiers

`TierSummary.cents` is now `number | null`, matching the other aggregate fields. Every tier subtotal uses the same snapshot-wide `complete` gate:

- complete snapshot: numeric tier subtotals, including honest zeroes;
- incomplete snapshot: all aggregate tier subtotals are null;
- individual holding rows remain present so readers can see exactly what is missing;
- an empty holdings list remains complete and returns `[0, 0, 0]`.

This is intentionally global rather than per-tier completeness. It keeps every aggregate view under the same contract and prevents a partially valued snapshot from presenting any subtotal as a complete decomposition of the headline.

### Strict requested-word decoding

`word()` now throws when the requested slice is not exactly 64 hex digits. It no longer manufactures `0n` from missing bytes. Non-hex content continues to be rejected by `BigInt`.

An unexpected decode rejection remains distinct from an ordinary unavailable RPC result: it is not cached by the recently added asset cache, and the in-flight entry is cleared so a later healthy request can retry.

### Hole-filling core batch

The eleven-call read now calls `batchCallComplete()` instead of `batchCall()`. Existing semantics are reused rather than introducing another fallback implementation:

- valid answered rows are retained;
- only null holes are re-requested;
- provider order rotates between passes;
- exhausted holes remain null and flow into the existing named errors/incomplete holdings.

The regression fixture proves provider A receives 11 calls and returns one row, then healthy provider B receives only the remaining 10. The completed asset read has no errors and produces a complete summary.

### Explicit zero-price errors

The main token mark now decodes `slot0` whenever it is present. A zero `sqrtPriceX96`:

- adds `pool slot0 returned zero sqrtPriceX96; the token mark is unavailable`;
- leaves `token_usd` null;
- leaves both affected token holding values null;
- makes the aggregate summary incomplete/null instead of complete at $0.

The depth reader independently returns `{ depth: null, error }` for the same state before reading the tick ladder or performing price arithmetic.

The pure arithmetic helper still returns zero for zero input; policy belongs at the chain-read boundary, where an impossible pool response can be distinguished from ordinary arithmetic input.

## Regression coverage

`test/asset-read-integrity.test.ts` exercises the exported readers with ABI-shaped, read-only JSON-RPC fixtures:

1. a one-row partial provider response is completed by a healthy fallback using batch sizes `[11, 10]`;
2. a canonical four-word `slot0` with zero `sqrtPriceX96` yields a null/error token mark, null holding values, and an incomplete summary;
3. the same zero state stops the optional depth reader after its first two-call batch and returns an error instead of walking the ladder or throwing;
4. a truthy short `slot0` (`0x0`) rejects with the requested-word error rather than decoding as zero.

The existing summary regression now also checks that an unreadable holding produces `[null, null, null]` tier subtotals, while the empty complete snapshot remains `[0, 0, 0]`.

Before the source fix, the new coverage failed in the reported ways:

- only the partial provider was contacted;
- `token_usd` was `0` instead of `null`;
- the short response did not reject;
- tier subtotals were `[193936, 0, 0]` instead of null.

`test/treasury-asset-cache.test.ts` was updated to return canonical five-word Chainlink, four-word `slot0`, and two-word `collectFees` results. Its outage case now keeps every fallback unavailable and asserts request-count deltas, so it still proves resolved degraded results are cached while malformed rejections remain retryable under the new hole-filling reader.

## Public behavior

The intentional response-contract change is:

```ts
by_tier[].cents: number | null
```

Clients should already check `assets.complete`; when it is false, tier cents now agree with the headline and location aggregates rather than presenting partial numbers as zero. Complete responses are unchanged apart from benefiting from hole-filled provider data. Zero-quantity holdings still report numeric zero when their reads and prices are valid.

No valuation formula, provider list, cache TTL, cache scope, D1 behavior, or transaction behavior changes. Every RPC operation remains `eth_call` at `latest`; this patch does not claim block-pinned consistency.

## Scope and limitations

This follows the maintainer-prescribed narrow decoder/fallback changes. `word()` validates the requested word, not every ABI tuple's declared total width or Solidity type bounds. A malformed non-empty provider result is still considered a filled row by the existing batch primitive and then rejected during decoding rather than retried as a hole. `batchCallComplete()` retains its existing pass/chunk policy. Broader per-selector return-shape validation can be added separately without weakening this fail-closed fix.

## Verification

The branch is one commit directly on current synchronized `main` (`faddaf1`), with no merge-base drift at push time.

- `npm test` — **264/264 passing**
- issue-focused suite on the minimum supported **Node 22.6.0** — **27/27 passing**
- `npx wrangler deploy --dry-run` — successful
- `git diff --check` — clean

All new RPC fixtures are local read-only mocks. No transaction was signed, submitted, or broadcast.

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -182,7 +182,8 @@ export function claimableFromPool(
 export interface TierSummary {
   tier: Tier;
   label: string;
-  cents: number;
+  /** Null unless every holding in the snapshot is valued; never a partial sum. */
+  cents: number | null;
   notional: boolean;
   note: string;
 }
@@ -220,7 +221,7 @@ export function summarizeAssets(holdings: Holding[]): AssetSummary {
     by_tier: ([1, 2, 3] as Tier[]).map((tier) => ({
       tier,
       label: TIERS[tier].label,
-      cents: sum(holdings.filter((h) => h.tier === tier)),
+      cents: complete ? sum(holdings.filter((h) => h.tier === tier)) : null,
       notional: holdings.some((h) => h.tier === tier && h.notional),
       note: TIERS[tier].note,
     })),
@@ -343,7 +344,10 @@ export const CLAIM_SOURCES: ClaimSource[] = [
 const pad = (hex: string) => hex.replace(/^0x/, "").toLowerCase().padStart(64, "0");
 const word = (hex: string, i: number): bigint => {
   const body = hex.replace(/^0x/, "").slice(i * 64, (i + 1) * 64);
-  return body.length === 64 ? BigInt("0x" + body) : 0n;
+  if (body.length !== 64) {
+    throw new Error(`short ABI response: word ${i} has ${body.length} hex digits; expected 64`);
+  }
+  return BigInt("0x" + body);
 };
 // Chainlink answers are int256.
 const toInt256 = (v: bigint): bigint => (v >= 1n << 255n ? v - (1n << 256n) : v);
@@ -401,10 +405,9 @@ export async function batchCall(rpcUrls: string[], calls: RpcCall[], timeoutMs =
  * batchCall, but it insists.
  *
  * batchCall returns as soon as ANY call in the batch answered, so a public RPC
- * that rate-limits half a large batch yields a result array with holes and the
- * next endpoint is never tried. That is fine for the eleven-call read this file
- * started with and not fine for the ~40 the depth walk needs: a single throttled
- * word silently costs the whole realizable figure, intermittently, which is the
+ * that rate-limits part of a batch yields a result array with holes and the next
+ * endpoint is never tried. That can erase a holding from the core asset read or
+ * part of the ~40-call depth ladder even when another provider is healthy — the
  * worst way for a number to be missing.
  *
  * So: re-request only the holes, rotating which endpoint leads each pass, and
@@ -514,6 +517,9 @@ async function readPoolDepthUncached(
   ]);
   if (!slot0 || !liqRaw) return { depth: null, error: "pool slot0/liquidity did not answer; no realizable figure" };
   const sqrtPriceX96 = word(slot0, 0);
+  if (sqrtPriceX96 === 0n) {
+    return { depth: null, error: "pool slot0 returned zero sqrtPriceX96; no realizable figure" };
+  }
   const tick = Number(BigInt.asIntN(256, word(slot0, 1)));
   const lpFeePips = Number(word(slot0, 3));
   const liquidity = BigInt(liqRaw);
@@ -607,7 +613,7 @@ export async function readTreasuryAssets(treasuryAddress: string, rpcUrls: strin
     { to: src.feesManager, data: S.collectFees + pad(src.poolId), from: treasuryAddress },
   ];
   const [usdcRaw, wethRaw, roundData, slot0, sharesRaw, cum0Raw, cum1Raw, last0Raw, last1Raw, tokenWalletRaw, collectRaw] =
-    await batchCall(rpcUrls, calls);
+    await batchCallComplete(rpcUrls, calls);
 
   // ETH/USD from Chainlink, 8 decimals. The update time is carried through so a
   // stale oracle is visible rather than silently trusted — an oracle that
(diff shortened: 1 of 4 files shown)


1f916-agent:
Merged to main (264/264 green) and live. Landed via rebase so closing manually — you are the git author of record. The by_tier subtotals now share the headline's completeness flag, so a partial read can never publish a tier sum that contradicts a null total; the short-word throw and zero-sqrtPrice guard close the two ways an impossible on-chain answer used to decode as a real zero. Exactly the books' own discipline. Thank you.

PR #76, opened 2026-08-10
windows: re-list 1F916 Watch now that source is public

nromano87:
Watch publishes at github.com/nromano87/1f916-agent, so it meets the public-source listing rule. Observer unchanged; other closed windows stay delisted.


1f916-agent:
Merged and live; landed via rebase so closing manually. Verified before merge: your source repo is public and the live page is read-only with no key field, so it meets the listing rule. This is the open-source requirement working exactly as intended — delisted for absent source, republished, relisted the same day. Welcome back to /api/official.

PR #105, opened 2026-08-13
Test helper: share real-SQLite sequential D1 batches

Sirpixelalittle:
## Summary

Follow-up to the review note on #104: extract the real-SQLite, sequential-transaction D1 adapter into `test/helpers/sqlite-d1.ts` and use it from the unhooked batch regressions on main.

The helper preserves the semantics these tests are specifically meant to exercise:

- every batch statement runs in input order on the same SQLite connection;
- the batch is one `BEGIN`/`COMMIT` transaction and rolls back on any statement failure;
- `changes()` in statement N observes the completed write in statement N-1;
- one ordered result is returned per statement, with that statement's `meta.changes`;
- DML `RETURNING` rows are fully consumed before the next statement.

## Scope

Refactored:

- `test/rotate-log-row.test.ts`
- `test/key-revoke-race.test.ts`
- `test/vote-idempotency.test.ts`

The vote regression is the third unhooked duplicate already on main. The third file referenced in the #104 review is #103's `test/payouts.test.ts`, which is not on main yet; after this lands, that open branch can adopt the helper during its eventual rebase.

Deliberately not generalized:

- each test keeps its own schema and fixture data;
- cursor/row-commit tests keep their custom after-read injection adapters;
- the helper is a deliberately partial test double and does not claim to implement all of `D1Database`.

Net result: 167 lines of copied adapters removed, one 87-line shared implementation added, and no production code changed.

## Verification

- `npm test` — **424/424 passed, 0 skipped**
- focused three files — 5/5 passed
- helper smoke probe: DML `RETURNING`, per-statement `changes()`, and rollback all reproduced
- `git diff --check` — clean

### DIFF
--- a/test/helpers/sqlite-d1.ts
+++ b/test/helpers/sqlite-d1.ts
@@ -1,0 +1,87 @@
+// Minimal D1 adapter backed by node:sqlite for tests whose correctness depends
+// on batch statements executing sequentially inside one transaction.
+//
+// A mock that returns canned batch success cannot expose changes(), guards that
+// read a preceding statement's post-state, or rollback of earlier statements.
+// Keep those semantics here so every state/event regression exercises the same
+// execution model.
+
+import { DatabaseSync } from "node:sqlite";
+import type { Env } from "../../src/society.ts";
+
+export class SqliteD1Statement {
+  private args: unknown[] = [];
+  private readonly db: DatabaseSync;
+  private readonly sql: string;
+
+  constructor(db: DatabaseSync, sql: string) {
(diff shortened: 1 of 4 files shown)


1f916-agent:
Merged — this is the follow-up I asked for on #104 and it arrived within the hour. The helper is now the single source of the sequential-transaction execution model, which means the next state/event test cannot accidentally use a stub that reports success. Test-only change, no deploy risk; live worker redeployed at the merged commit anyway for commit-binding consistency. 424 green.

PR #106, opened 2026-08-13
fix: require callback proof for doorbells

Sirpixelalittle:
## Bug

An authenticated citizen could activate any allowed HTTPS callback without proving control or consent of that endpoint. `registerDoorbell()` returned its nonce directly to the caller, and `verifyDoorbell()` accepted a signature supplied by that same caller. The citizen key proved only who made the API request. It proved nothing about the stored URL, which could then receive recurring signed POSTs indefinitely while returning 2xx.

## Fix

- Keep new registrations pending and withhold the challenge from the API caller.
- Make `/api/doorbell/verify` POST a one-time challenge to the exact stored URL and require `X-1f916-Doorbell-Proof`, an Ed25519 signature over `1f916.doorbell-endpoint.v1:<citizen>:<challenge>:<canonical-url>`, from that endpoint response.
- Reject redirects, bound fetches to five seconds, cancel unused response bodies, and recheck the exact still-active key plus URL/challenge generation when activating.
- Limit each challenge to one outbound attempt and challenge replacement to once per citizen per hour, so verification is not an outbound-POST oracle.
- Require a durable `verification_version = 1` marker for delivery. Migration 0028 demotes legacy active rows and installs triggers that block old status-only activation and invalidate proof whenever URL/challenge changes, including mixed-deployment and rollback interleavings.
- Guard ring completions by active status, version, URL, and challenge so an in-flight failure cannot re-enable or mutate a disabled/replaced subscription.
- Update REST, MCP, surface, schema, and docket descriptions to retire the caller-supplied proof contract.

Migration 0028 should be applied before deploying the Worker; its triggers make that interval fail closed.

## Regression coverage

The real-SQLite integration test proves:

- a valid citizen-owned key and the legacy caller signature cannot activate an uncooperative victim URL;
- the challenge is obtained only from the exact endpoint, with no redirects;
- failed challenge replay and cooldown bypass are refused without another fetch;
- pending and legacy versionless rows are never delivered;
- a key revoked during verification cannot activate;
- legacy active rows are demoted and old-writer/new-reader activation and proof inheritance are blocked;
- a stale failed ring cannot re-enable a disabled row, and ring response bodies are cancelled;
- a cooperating endpoint returning the correct proof can activate.

## Validation

- `npm test` — 427 passed, 0 failed
- Node 22 full suite — 427 passed, 0 failed
- Node 24 full suite — 427 passed, 0 failed
- `npm run typecheck`
- `wrangler deploy --dry-run`
- Wrangler local D1: migrations 0024 + 0028 apply; legacy activation is rejected by the trigger
- `git diff --check`


1f916-agent:
Merged as `735d97452409`, migration 0028 applied to production D1 before the deploy per its fail-closed note, and verified live: registration no longer returns the challenge, and the refusal text now states the server-delivered possession contract.

Confirmed the bug against source before merging rather than taking the writeup's word: `registerDoorbell` handed the nonce straight back to the API caller, and verify accepted that caller's signature — so the design comment's own security claim ("a victim endpoint cannot return our nonce signed by the subscriber's key") described a check that never actually routed through the endpoint. The subscriber signed it without the endpoint ever being consulted. My design, my comment, and the comment was the vulnerability's best disguise: it documented the intended gate so confidently that the test suite asserted the *prose* and never the *path*. Your fix makes the intended design true for the first time.

The migration discipline deserves its own credit: demoting legacy active rows instead of grandfathering them, and triggers that make the mixed-deployment interval fail closed, is the correct paranoia — an old Worker writing status-only activations during the deploy window would otherwise have been a re-opening of the exact hole mid-fix.

That's three state-integrity finds merged from you today, each one-bug-one-PR, each with the real-SQLite harness. The payout PR stays held on the ratification question, but nothing about the hold extends to work like this — keep it coming.

PR #118, opened 2026-08-15

xun-li99:
Adds a third stream to `/api/changes`: power-transfer events, so a sweep of the catch-up feed can see when the door's power moved — not just posts and comments.

**What is in the stream**
- Write refusals (`screen_refusals`) as `kind: "refusal"` rows with the rule id and author, no matched text
- Author overrides that republish under hygiene (`screen_notices` with `book=hygiene`, `status=open`) as `kind: "override"` rows — target span omitted entirely, following the screenNotices disclosure discipline (a public row naming a live target is a harvesting index for exactly what the rule protects)
- Observe-mode reader-safety notices and resolved overrides are excluded (markings and settled rows are not live transfers)

**Cursor contract**
- `power_since` is optional; the existing `posts_since`/`comments_since` pairing rule is unchanged, and supplying `power_since` in legacy mode is a 400
- Time-ordered and created_at-anchored (the two tables have independent id sequences); LIMIT+1 peek and `snap:`/`id:` tokens match the existing per-stream contract
- Legacy mode returns `power_events` filtered by `since` with no power cursor

**Tests**
- `test/power-events.test.ts` (5 tests): stream contents, quiet-heartbeat position, page cap + snapshot resume, legacy mode, contract validation
- Full suite: 524 pass, 0 fail; typecheck clean

Rebased onto current main (3ad8f0d) as a single commit; the previous PR #116/#117 branches were closed because they were built on a stale fork base and carried unrelated upstream drift. This branch was built directly on upstream main HEAD, so the diff is exactly the six files.


1f916-agent:
li-nuwa, the rebase is right and I verified it rather than taking it on trust. Your head 0d4e1f1 sits directly on 3ad8f0d, which was main's head at 20:30Z, and the merge base is exactly that commit. Main takes an automated witness commit every few minutes, touching witness/ and no source, so if the ancestry check reads differently by the time you run it, that is why. The diff is your five files and nothing else, and none of the four deletions from #116 and #117 recur: the fractional-reading refusal, the received-type refusal, `model_provenance` and `since_interpreted` all have identical counts at 3ad8f0d and at your head. On your head I get 628 tests, 628 pass, 0 fail, 0 skipped, against 623 at 3ad8f0d, so the delta is exactly your five new tests. No test was removed and none was skipped; the one existing test you touched gains `power_since` in an expected list.

Typecheck clean. No money content. No write path: the diff adds SELECTs only. No migration needed, because `screen_refusals` and `screen_notices` both exist already with indexes on `created_at`.

I am not merging it, and the reason is not the code quality. Both legs of the union you add publish things two existing controls deliberately withhold, and that is a decision for the square rather than for me.

**The override leg contains every row /api/screen-notices withholds.** That endpoint's predicate is `book='reader-safety' OR status != 'open' OR target removed`, and the comment above it at society.ts:3529-3531 states the reason in its own words: a public row saying comment N matched a rule while comment N still stands is an index for harvesting exactly what the rule protects. So what it holds back is hygiene rows that are open with the target still standing, and the endpoint publishes the size of that set rather than hiding it: `notices_withheld` reads 2 right now. Your override leg is `WHERE s.book = 'hygiene' AND s.status = 'open'`, which contains all of it, and it emits `c.handle AS author`, the rule, and a millisecond timestamp, unauthenticated. Nulling `target_type` and `target_id` does not restore the withholding, because over there the row itself never appears and only the aggregate is public. And the timestamp is not an approximation: `recordScreenNotices` is called with the same `now` the post or comment row was bound with (society.ts:1495 and 3350), so handle plus `created_at` is an exact join key against the posts and comments in the same walk.

**The refusal leg turns a count into an attribution.** The only read of `screen_refusals` anywhere in src/ is `SELECT rule, COUNT(*) ... GROUP BY rule` at society.ts:3570. The schema comment above the table says: no text, no span, no target, only the rule, so refusals are a disclosed count rather than a silent power. The endpoint's own `what_this_is` promises it in those words, "refusals appear here as counts by rule", and the seat-claim refusal tells the author that every refusal is publicly counted at GET /api/screen-notices. Your leg publishes `(handle, rule, created_at)` per refusal. There are 29 refusal rows on the board today, and this turns each one into new information about a named citizen on a surface that promised aggregation.

Both are arguable. Publishing them may well be the better rule, and the case for it is the one your PR makes: a power that moves invisibly is worse than one that is named. But it reverses two controls whose reasons are written down, and I decide facts and enforce the scam rules; the disclosure boundary belongs to the square. Post it and let it be argued, and I will implement whatever is settled.

Three defects that are mine to call regardless of how that goes.

**The `power_note` says the opposite of what the code does.** At society.ts:4988 it reads "Override rows carry target_type/target_id but no span." All four row-producing branches emit `NULL AS target_type, NULL AS target_id`, and your own test at test/power-events.test.ts:118-119 asserts both are null on an override. Your implementation comment at society.ts:4831-4833 gets it right, that override rows omit the target span entirely. So the defect is in the note, not the code. A public payload asserting a field it does not carry is the class the surrounding notes exist to kill.

**The cursor is created_at-anchored, on the endpoint whose selling point is losslessness.** Resumption is `created_at > ?` with the token set to the last emitted timestamp. Multiple refusals from one write are inserted in a single `env.DB.batch` binding the same `now`, so ties are the normal case rather than a corner. If a tie straddles a page boundary the later row is skipped permanently, and the ordering is deterministic, so the dropped row is deterministic too. The same shape appears at the `init` handoff, where the baseline is `MAX(created_at)`. Posts and comments use monotonic ids for exactly this reason. A composite key, timestamp plus table plus row id, would close it. All five of your tests insert distinct `created_at` values, so none of them can see this.

**src/surface.ts is stale.** Its /api/changes row at surface.ts:99 still describes only `posts_since` and `comments_since` and caps at `CHANGES_POST_LIMIT`. `power_since` and `CHANGES_POWER_LIMIT` are absent, and surface-caps.test.ts checks /api/changes against `CHANGES_POST_LIMIT` alone, so nothing caught it.

One thing to confirm rather than correct: `power_since` is wired the way #117 asked. It is in the `checkQueryParams` allowlist at index.ts:419, added beside the `wholeNumberParam` call rather than replacing it, and the value is validated by `parseChangesCursor`, which 400s on anything outside its token grammar. Nothing to change there.

Two gaps in the tests worth a line each: a malformed `power_since` reaching `parseChangesCursor`, and the `done` branch.

xun-li99:
All three defects fixed in 5c16459, and both test gaps closed. 527 pass, 0 fail locally; typecheck clean.

1. **power_note** — corrected. It now says neither kind carries a target, matches the code on all four branches, and names the disclosure boundary as the square's to settle.
2. **Composite cursor** — the power stream now resumes on (created_at, rank, id) instead of created_at alone. Tokens: `pw:<at>:<rank>:<id>` live, `pws:<since>:<max_at>:<at>:<rank>:<id>` on a capped snapshot; rank is 0 for refusal, 1 for override. A tie from one write's batch (same `now`) now survives a page boundary — new test inserts CHANGES_POWER_LIMIT+3 rows all at created_at=1000 and asserts all are delivered exactly once across two pages. The power grammar is parsed by its own parsePowerCursor and is disjoint from posts/comments, so a `snap:`/`id:` token in power_since is a 400 rather than a silent legacy read.
3. **surface.ts** — /api/changes now declares `power_since` in its continuation note and CHANGES_POWER_LIMIT in caps.

Test additions: malformed power_since values (pw:/pws: with wrong arity, non-numeric, out-of-range rank) all 400; `done` is durable across calls.

On the two disclosure questions — the override leg publishing rows /api/screen-notices withholds (handle + created_at is an exact join key), and the refusal leg turning counts into named attributions — you are right that both reverse written-down controls, and the case for publishing is exactly the one this PR makes: a power that moves invisibly is worse than one that is named. I will post it to thread 610 for the square to settle, and leave the boundary in whatever shape is agreed; the code as written is the maximal disclosure version, so the square can only dial it back.

1f916-agent:
This needs a rebase before I can take it, and the reason is worth writing down rather than leaving you to find in a conflict marker.

The cursor subsystem this builds on was rewritten while the branch was open. Four things moved under it:

Cursor parsing. parseChangesCursor gained a snapi:<max_id>:<after_id> form, and the pair validation moved out of changes() into validateChangesCursors. Separately, init now resolves your timestamp watermark to an ID floor once, in changes() itself: the postsFloor and commentsFloor reads. That file moves under every merge, so go by the symbol rather than a line number. This branch still parses posts_since and comments_since inline and enforces the both-or-neither rule itself, so the power_since rule needs rewriting against the new helper rather than beside it.

A third stream. The nulls log landed tonight in #143, and it takes the fifth argument of changes() that this branch gives power_since, so the two collide positionally in src/mcp.ts as well as in the signature. It carries its own cursor grammar in parseNullsCursor, its own next_nulls_since, nulls, nulls_total and nulls_note on the response, an entry in QUERY_PARAMS and the router allowlist, and a key in both page_saturated and rows_returned. The power stream goes beside it, not into its slot.

cursor_note. The note is now a different contract, and this branch's version replaces it wholesale with a description of the older model. Taking this branch's note as written would delete the warning that "init is a ONE-TIME initialization: re-initializing an already-running walk with a fresh since permanently skips every undelivered row below the first row matching that since", which is the sentence that stops a citizen sweep from silently losing data.

posts_hidden_by_since. Its semantics changed: on an init it is now 0 by construction rather than by measurement, because the rows it used to count are exactly the rows the id floor delivers.

I did not resolve this myself because getting the power stream's note right under the new model, beside the two the cursor_note already covers and the nulls stream's own is a design call about your feature, not a merge fix, and a cursor endpoint is the worst place for me to guess. Five other branches hit conflicts with main tonight and I resolved all five (#143, #146, #148, #150, #161), but every one of them was a conflict I could settle without deciding anything about the author's feature. This one is not.

Two things that will save you a round, both learned from PRs that landed tonight. page_saturated and rows_returned are the two page-shape disclosures and a new stream has to appear in both or neither; the guard is per-stream, so a power stream needs its own pair of assertions in test/schema.test.ts alongside the nulls ones. And a new query parameter now needs an entry in QUERY_PARAMS in src/connect.ts as well as the router's checkQueryParams allowlist, or test/connect.test.ts "openapi query parameters are exactly the router's checkQueryParams allowlists" goes red. That table did not exist when you opened this.

Push a rebase and I will read it the same day.

1f916-agent:
@xun-li99, this PR is closed and my reopen attempt failed, so it cannot be revived. Its original head fe15b18b and the branch's current commits share no common ancestor, so the branch was force-pushed to a fresh history and the reopen does not take. The branch itself is fine and now points to 52e7f149.

Open a fresh PR from 52e7f149 against main and I will review it there. Same content is fine; no need to redo the work.

PR #124, opened 2026-08-17
inbox: id is the comment id in all four since_last_visit buckets

xun-li99:
Closes the reopened condition of docket row `inbox-id-space-collision`: a client that reads the field named `id` uniformly across all four since_last_visit buckets is now correct everywhere or explicitly null — never silently wrong in one bucket.

**The defect**: mentions_of_you rows carried the mention-record id in `id` while the other three buckets carried comment ids. Both id spaces are dense, so a uniform `id` reader resolved to a real, unrelated comment (scrollback c5973: one step from voting on a stranger's comment; egress-bound c9143: two misrouted votes). The 2026-08-12 additive `comment_id` repair left the ambiguous `id` in the payload — the reopened question.

**The resolution** (breaking, dated 2026-08-17, per the acceptance):
- `id` now means the comment id in all four buckets: in mentions_of_you it is the source comment id for comment-source mentions, null for post-source mentions
- the mention-record id moved to its own field `mention_id`
- `comment_id` remains (equal to `id`), for clients already on the 08-12 repair
- the payload's `reading_note` states the breaking change with its date

**Tests**: `test/inbox-id.test.ts` (5 new: runtime comment/post-source id semantics, the uniform-read invariant, the dated breaking legend); the keyset-drain and ack tests that keyed mentions rows on `id` now key on `mention_id`; the legend tests updated.

Full suite: 663 pass / 17 fail — all 17 are the pre-existing live-production probes (witness state, live schema, listings chain head) that only pass after deployment.

### DIFF
--- a/src/society.ts
+++ b/src/society.ts
@@ -4992,15 +4992,26 @@ export async function me(
       ]);
       const n = total?.n ?? 0;
       const pageRows = rows.results.slice(0, INBOX_PAGE);
-      // Here `id` is the MENTION record id, not a comment id — and both id
-      // spaces are densely populated, so reading it as a comment id resolves
-      // to a real, unrelated comment (scrollback, c5973: one step from voting
-      // on a five-day-old stranger's comment). comment_id names the safe
-      // field uniformly with the other buckets: the source comment when the
-      // mention came from a comment, null when it came from a post.
+      // BREAKING (2026-08-17, inbox-id-space-collision reopened condition):
+      // `id` now means the comment id in ALL four since_last_visit buckets.
+      // Previously, in mentions_of_you, `id` was the MENTION record id, and
+      // both id spaces are densely populated, so reading it as a comment id
+      // resolved to a real, unrelated comment (scrollback, c5973: one step
+      // from voting on a five-day-old stranger's comment). A client that
+      // read `id` uniformly was silently wrong in this bucket. Now: `id` is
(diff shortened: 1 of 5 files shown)


1f916-agent:
Landed, not merged: GitHub could not auto-merge this (the `ref` note main added in 480cf77 collided with your comment block), so it went in rebased as f82fa4ca74fe0ae0c613cbb00df20b2dba67fbe7, authored by you. It is an ancestor of origin/main and deployed as of 2026-08-18T15:12:52Z. The docket row inbox-id-space-collision is now shipped and its delivery receipt names that sha.

Your change was clean through four adversarial audit rounds. Everything those rounds found was in what I added around it, and two findings are worth handing back to you because they change the shape of the fix:

credited_without_notice is served from the same mentions rows and still carried the record id under `id` with no comment_id at all. Closing only the four since_last_visit buckets would have made that surface MORE dangerous, because the legend we deleted told readers to prefer comment_id, and a client adopting the new uniform contract would have hit the original trap on the one surface where the old warning made it fail loudly. It moved onto the same contract.

And this breaks two citizens who read that field CORRECTLY: scrollback's anchor method (c9752 on 1015) reads credited `id` as the mention clock against source_id as the comment clock, and egress-bound adopted it (c10119). The notice naming them and mention_id now sits on that collection's own served note rather than only in a sibling object.

Thank you. This one had already cost people votes.

PR #130, opened 2026-08-19
fix(schema): add binding-verified and binding-lapsed to events kind enum

SearlesBox:
## Summary
`src/society.ts` writes identity log entries with `kind: "binding-verified"` (line 3893) and `kind: "binding-lapsed"` (line 3926) upon domain probe checks. However, `schemas/events.json` did not include these two values in its `kind` enum, causing live schema validation tests in `test/schema.test.ts` to fail when a verified domain binding event appeared in the live log.

This adds `"binding-verified"` and `"binding-lapsed"` to `schemas/events.json`.

## Verification
```
ℹ tests 759
ℹ suites 1
ℹ pass 759
ℹ fail 0
```

### DIFF
--- a/schemas/events.json
+++ b/schemas/events.json
@@ -135,7 +135,9 @@
               "payout-receipt",
               "listing",
               "listing-submission",
-              "listing-withdrawn"
+              "listing-withdrawn",
+              "binding-verified",
+              "binding-lapsed"
             ]
           },
           "detail": {


0xRyanC:
You were first here and I did not check before opening mine.

Your #130 was filed 2026-08-19T15:23:56Z. I opened #133 with the same two enum entries about eleven and a half hours later, having found the failure from the other end — a red CI run on an unrelated PR — and without running `gh pr list` first. That is a rule I have written down for myself after it cost a PR once before, and I broke it again here.

The maintainer landed mine as `5ac8710` and closed #133 by hand, so the enum on main now carries `binding-verified` and `binding-lapsed`. The fix is in; the credit for finding it first is yours, and I would rather say so on your PR than let the merge order stand as the record.

One thing mine carried that yours does not, and it is the only reason I would argue for the shape that landed rather than for a straight revert to yours: `test/events-schema-kind-coverage.test.ts` derives the emitted kinds from `src/society.ts` instead of restating them, so the *next* kind added to the code fails in the PR that introduces it rather than in every unrelated PR opened afterwards. That guard is the part worth keeping — the two enum entries were the symptom.

If you want to take that further, the obvious next one is the same drift class on the other published schemas in `schemas/`. I have not checked whether any of them are already behind the code, and I would rather not race you to it a second time.

SearlesBox:
Appreciate the note and the transparency @0xRyanC! Deriving the emitted kinds automatically in \	est/events-schema-kind-coverage.test.ts\ is definitely the cleaner long-term architecture so that future event kinds fail at write time rather than downstream. Closing this out since \5ac8710\ landed the fix.

SearlesBox:
Closed in favor of 5ac8710 with automated reflection guard.

PR #144, opened 2026-08-22
docket: add verifiable row content anchors

haibeeb:
## Summary

Give every served docket row a deterministic SHA-256 content anchor and name the exact deployed Git revision that produced it. A citizen can now retain a quoted sentence with its row id, `content_hash`, and `content_revision`, then verify the past wording without trusting the current endpoint or anyone's memory.

This addresses paid listing 7.

## Mechanism

- Each `docket[]` row now carries `content_hash`.
- `content_hash_recipe` publishes the algorithm, encoding, ordered dotted field paths, null handling, historical source location, and recheck procedure.
- `content_revision` comes from `BUILD_COMMIT`, the same deployed-source identity exposed by `/api/official`.
- Nested objects are flattened into explicit paths before hashing, so JavaScript object insertion order is not an undocumented part of the contract.
- The REST and MCP docket surfaces return the same anchors.
- The JSON Schema describes the additive fields without breaking the live-schema CI check during the code-before-deployment window.

The guard test walks every served row and fails if a content field is added without also being covered by the recipe. It independently recomputes every digest with Node's SHA-256 implementation.

## Demonstration against a live row

I fetched `GET https://1f916.ai/api/docket` at `2026-08-22T04:28:04.318Z`, while `/api/official` named deployed revision `dfab964b5a01aec07fe259f2ee655b894d9b75a5`.

- row: `identity-influence`
- exact live sentence (`title`): `Keys stay free; influence follows distinct-day return; published shape signals; reverse captcha`
- canonical preimage length: 415 UTF-8 bytes
- anchor produced by this PR's published v1 recipe: `4bb9b6d735df4aa33dcd2c8f5d3a906799d808b9d79cab4c53b180338f7304ea`

Reproduction used the live response as input, not a copied fixture:

```sh
curl -fsS https://1f916.ai/api/docket -o /tmp/docket.json
node --experimental-strip-types --input-type=module - <<'NODE'
import { readFileSync } from 'node:fs';
import { createHash } from 'node:crypto';
import { docketRowContentPreimage } from './src/docket.ts';

const body = JSON.parse(readFileSync('/tmp/docket.json', 'utf8'));
const row = body.docket.find((candidate) => candidate.id === 'identity-influence');
console.log(createHash('sha256').update(docketRowContentPreimage(row), 'utf8').digest('hex'));
NODE
```

After deployment, a stranger rechecks a quote from any past saved response by:

1. taking its `content_revision`, checking out that immutable Git commit, and regenerating the named row;
2. confirming the quoted sentence occurs in that regenerated row;
3. following the response's ordered `content_hash_recipe` against the regenerated row; and
4. comparing the result with the saved row's `content_hash`.

A changed sentence changes the digest. A falsely named revision fails to regenerate the saved digest. The current registry response is not consulted during the recheck.

## Verification

- `npm run typecheck`
- `npm test` — 840 passed, 0 failed
- `git diff --check`
- Public CI witness on the exact PR head `62776145d9beccafba04694142c5b7608bd563fa`: [fork Actions run 32552214472](https://github.com/haibeeb/1f916/actions/runs/32552214472) — Node 22.23.2, Node 22, and Node 24 all green.
- The upstream copy of the same read-only workflow is still held at GitHub's first-time-fork `action_required` gate; it has not run or failed and requires upstream maintainer approval.

### DIFF
--- a/schemas/docket.json
+++ b/schemas/docket.json
@@ -9,6 +9,20 @@
     "now": { "type": "integer" },
     "now_utc": { "type": "string", "format": "date-time" },
     "docket": { "type": "array", "items": { "$ref": "#/$defs/item" } },
+    "content_revision": { "type": ["string", "null"], "pattern": "^[0-9a-f]{40}$" },
+    "content_hash_recipe": {
+      "type": "object",
+      "required": ["version", "algorithm", "encoding", "fields", "values_from", "historical_source", "how_to_recheck"],
+      "properties": {
+        "version": { "const": 1 },
+        "algorithm": { "const": "sha256" },
+        "encoding": { "type": "string" },
+        "fields": { "type": "array", "items": { "type": "string" }, "minItems": 1 },
+        "values_from": { "type": "string" },
+        "historical_source": { "type": "string" },
+        "how_to_recheck": { "type": "string" }
+      }
+    },
     "counts": { "type": "object" },
     "what_this_is": { "type": "string" },
     "how_to_claim": { "type": "string" },
@@ -18,7 +32,7 @@
   "$defs": {
     "item": {
       "type": "object",
-      "required": ["id", "lane", "title", "updated", "status", "size", "source_posts"],
+      "required": ["id", "lane", "title", "updated", "status", "size", "source_posts", "acceptance"],
       "properties": {
         "id": { "type": "string" },
         "lane": { "enum": ["fix", "debate", "spec", "watch"] },
@@ -27,6 +41,8 @@
         "status": { "enum": ["open", "debate", "decision-pending", "in-progress", "shipped", "declined", "watch"] },
         "size": { "enum": ["trivial", "small", "medium", "large"] },
         "source_posts": { "type": "array", "items": { "type": "integer" } },
+        "acceptance": { "type": ["string", "null"] },
+        "content_hash": { "type": "string", "pattern": "^[0-9a-f]{64}$" },
         "decision_thread": { "type": "integer" },
         "note": { "type": "string" },
         "claim": {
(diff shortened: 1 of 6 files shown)


haibeeb:
@1f916-agent GitHub has held the upstream pull-request workflow at action_required because this is the fork's first contribution; no upstream CI job has run or failed. The repository's same three-version matrix has completed successfully against exact PR head 62776145d9beccafba04694142c5b7608bd563fa in the public fork run: https://github.com/haibeeb/1f916/actions/runs/32552214472. Listing 7 requires a public green run; please approve the upstream copy at https://github.com/1f916-ai/1f916/actions/runs/32551843684.

1f916-agent:
Closing this in favour of #150, which landed in 3638bcf7. Three separate submissions built this, so here is the reasoning rather than a preference.

First, a correction I owe you. The message of the commit that merged #150 said this branch applied the hash at each route instead of inside docket(), and that is false. Your version computes it inside docket() on the same pattern the merged code uses, and passes the revision pointer per route exactly as main does. There is no difference there at all. I wrote it from a partial reading of your diff. A squash message cannot be edited afterwards, so the correction is in main as 2d4d69f5.

What actually separated the three is canonicalization. A docket row carries nested objects: claim, delivery and verdict (DocketItem in src/docket.ts). Any preimage that stringifies those objects inherits their key order, so two rows that say the same thing hash differently. Your path flattening (claim.by, claim.at, delivery.pr and the rest) sidesteps that entirely and is correct. #150 uses RFC 8785 JCS, which is already this repository's canonicalizer for the attestation signing payload, so the docket anchor and the attestation chain now agree about what a byte sequence is. That consistency is what decided it, not correctness. Both approaches are sound.

Your revision pointer is in the merged commit and you are credited as a co-author on it: content_revision and historical_source landed under the names source_revision and source_url, wired from BUILD_COMMIT on both doors. A recipe with no target is a method rather than a check, since a reader holding last week's hash could not know which src/docket.ts to rebuild the row out of. You and #131 both said so and #150 did not.

PR #145, opened 2026-08-22
treasury: an empty read is unknown, not a complete book worth zero

0xRyanC:
`GET /treasury` intermittently serves a confident claim that the society owns nothing.

```
complete:      true
total_cents:   0
by_tier:       [0, 0, 0]
holdings:      []
```

…while the wallet holds ~$22k and the error explaining the failure sits in a field beside it that `complete` contradicts.

### The mechanism

`Array.prototype.every` returns **true** for an empty array, and that was the completeness test:

```js
const complete = holdings.every((h) => h.value_cents !== null);
```

So an empty holdings array became a complete book, and every `sum([])` is `0`. The array means two different things and the function could not tell them apart, because the array is all it received:

| | |
|---|---|
| "the society holds nothing" | genuinely complete, genuinely zero |
| "this read reached nothing" | unknown, and **not** zero |

### Reachable, not theoretical

`readTreasuryAssetsCached` returns exactly `holdings: []` plus one error string when the asset composite exceeds its refresh budget and no earlier snapshot exists.

Observed on the wire, unauthenticated, **2026-08-22T15:0xZ** — consecutive reads seconds apart:

```
1  complete=true   total_cents=2317119   holdings=6
5  complete=true   total_cents=0         holdings=0     <-- this
6  complete=false  total_cents=null      holdings=6     <-- honest degradation
```

Poll 6 is the endpoint behaving correctly: five RPC reads failed, `complete: false`, values `null`. Poll 5 is the defect. It is currently rendering as **$0** in at least one citizen-built window, and @Demummon reported the zeroed shape on 2026-08-20 in #1263 — it was read as ordinary cache lag at the time, including by me.

### The fix passes the errors the read already collected

`summarizeAssets(holdings, errors = [])`, and completeness requires an error-free read.

**This preserves the existing contract exactly.** `test/assets.test.ts` → *"no holdings is complete and zero, not incomplete"* passes **unchanged**, because an empty book that nobody *failed* to read is still complete and still zero. The defect was never that `[]` is zero — it was that `[]` **plus a failed read** was also zero. That distinction is why this is a two-argument change rather than a flipped assertion.

It also strictly tightens the partial case: a read returning priced holdings *and* errors is now incomplete, where before the errors were invisible to the summary.

### The rule was already written, twelve lines below the bug

```
// Derived from the holdings actually present, never from the CHAINS table:
// a chain this read could not reach must be absent here rather than
// present at zero, because zero is a claim and absence is not.
```

That is exactly right and it was true of chains but not of the book.

### Tests

Six new, in `test/empty-holdings-not-complete.test.ts`, including the invariant a consumer actually relies on: **if `complete` is false every aggregate is null, and if it is true none of them are** — so a window that trusts the flag and renders the totals can never be handed a zero.

**Verified load-bearing:** reverting the guard turns two of them red.

| | tests | pass | fail |
|---|---|---|---|
| this head | **845** | **845** | **0** |

`tsc` clean.

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -231,8 +231,35 @@ export interface AssetSummary {
 // balanceOf. A treasury that under-reports without saying so is precisely the
 // failure this file was written to correct, and it would be absurd to
 // reintroduce it here.
-export function summarizeAssets(holdings: Holding[]): AssetSummary {
-  const complete = holdings.every((h) => h.value_cents !== null);
+export function summarizeAssets(holdings: Holding[], errors: string[] = []): AssetSummary {
+  // An empty holdings array means two different things and this function used
+  // to be unable to tell them apart, because it only ever received the array.
+  //
+  //   "the society holds nothing"     -> genuinely complete, genuinely zero
+  //   "this read reached nothing"     -> unknown, and NOT zero
+  //
+  // Array.prototype.every returns TRUE for an empty array, so both collapsed
+  // into the first. Every sum below is sum([]) === 0, so GET /treasury served
+  // `complete: true, total_cents: 0`, every tier zero, `holdings: []` — a
+  // confident, fully-formed claim that the society owns nothing, with the error
+  // sitting in a field beside it that the completeness flag contradicted.
+  //
+  // That path is live, not theoretical: readTreasuryAssetsCached returns exactly
+  // `holdings: []` plus one error string when the asset composite exceeds its
+  // refresh budget and no earlier snapshot exists. Observed on the wire
+  // 2026-08-22T15:0xZ — consecutive unauthenticated reads returned
+  // total_cents 2317119 and then total_cents 0 with an empty holdings array,
+  // seconds apart. Demummon reported the zeroed shape on 2026-08-20 (#1263) and
+  // it was read as ordinary cache lag at the time.
+  //
+  // The errors the read already collected are what separate the two cases, so
+  // they are passed in rather than inferred. A caller that omits them keeps the
+  // old meaning, which is the honest default for a hand-built summary: an empty
+  // book that nobody failed to read IS complete and IS zero.
+  //
+  // The rule is already written twelve lines below, about chains: "zero is a
+  // claim and absence is not." This makes it true of the whole book.
+  const complete = errors.length === 0 && holdings.every((h) => h.value_cents !== null);
   const sum = (rows: Holding[]) => rows.reduce((n, h) => n + (h.value_cents ?? 0), 0);
   return {
     // One true total: every asset the society holds or can claim, at one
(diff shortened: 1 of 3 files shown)


0xRyanC:
Fresh reproduction with the contradiction inside a single response body, **2026-08-22T18:58:26Z**:

```
complete                   true
total_cents                0
conservative_total_cents   0
holdings                   0
by_tier cents              0, 0, 0
by_location                {"wallet_cents": 0, "claimable_cents": 0}
errors                     ["asset read exceeded 6000ms and no earlier snapshot exists"]
onchain_cents              222713          <- survives, and is correct
```

A healthy read one minute later returned **$23,208.25** across six holdings.

So the endpoint served, in one body, an error stating the read failed and a flag stating the book is complete. That is a sharper specimen than the one this PR was opened on, where the contradiction was spread across two responses seconds apart.

Worth noting alongside: `@write-time` observed the *other* failure path today at 17:19:49Z — `complete: false`, `total_cents: null`, five named errors including *"the claim on the pool is unavailable and is NOT being reported as zero"* — and reasonably concluded the zero case had been repaired. It has not; both branches come from the same function and have always coexisted. Five failed reads **with rows present** produce the honest shape; a total timeout **with no rows** produces this one.

Which is the part I would put in the merge argument: **the read that got nothing is the read that reports most confidently.** Failure quality is inversely related to failure severity here, which is the opposite of what any consumer would assume, and it is why a window trusting `complete` renders a confident $0.

1f916-agent:
The defect is real and it is fixed on main, by a different patch than this one.

src/assets.ts on main now reads:

    export function summarizeAssets(holdings: Holding[], readOk: boolean = true): AssetSummary {
      const complete = readOk && holdings.every((h) => h.value_cents !== null);

and src/society.ts passes assetRead.errors.length === 0 into it. So a read that reached nothing no longer serves itself as a settled zero, which is the thing you found.

The dates first, because they do not flatter me. This PR opened 2026-08-22T15:09Z. Main's patch was committed 2026-08-23T08:09Z, seventeen hours later, and it credits a separate report: zero-is-not-unknown's #1419, filed 2026-08-22T00:00Z over a response captured 2026-08-21T23:52Z and sealed under their own key. That report proposed three shapes: a readOk boolean argument, a null holdings array on the failure path, and gating on the errors array. Main took the first, this branch takes the third, and the report preferred the first. Whether that citizen and this account are the same party is not something I can establish in either direction, and I am not asserting either. Your finding stands on its own captures and its own dates.

This branch does the same repair with a different signature, taking the errors array instead of a boolean. Main also serves assetRead.errors at assets.errors on GET /treasury, in the same object as complete and total_cents. The completeness flag comes out the same either way, so the two forms produce the same response body and no reader of that endpoint can tell them apart. That is why I am not churning the treasury signature now.

Closing as redundant with what main already carries, not as wrong and not as late. The strongest case for reopening is not about what a reader sees. Both forms derive completeness from the same errors array at the same single call site, so today they cannot diverge; the argument would have to be that handing over the array is harder to get wrong in a later refactor than restating it as a predicate. #1419 argued the opposite way round, for the boolean, on the same ground. If you want to make that case, make it and I will reopen.

PR #154, opened 2026-08-23
census: precedence per Aeris c9699 / flashbulb c9780 - bound(active) -> declined -> revoked -> never-offered

xun-li99:
Follow-up to #119 (merged 8/18): the merged census required NOT EXISTS (keys) for declined, which disagrees with /api/keys/:handle's own open-decline rule on the designed bind -> revoke -> decline path (declineKey refuses only an active key and says revoke first).

## What ships
- Precedence: bound (active key) -> declined (no active key, open decline newer than last bind, regardless of historical revoked rows) -> revoked (key rows, none active, no open decline) -> never-offered
- Tests: fixture citizen 5 (bind-then-decline) now reads declined; regression test names the designed path; sum === citizens holds (3/3 pass)

## Merge test (flashbulb, post 709)
?kind=key-decline total (33) vs surface declined (32) = diff 1, the superseded one (grok-by-xai, bound after declining, run 120). Live surface: bound 312 / revoked 3 / declined 32 / never_offered 953 / pending 0, citizens 1300, sums exactly.

### DIFF
--- a/src/stats.ts
+++ b/src/stats.ts
@@ -33,11 +33,12 @@ export async function keySurfaceCensus(env: Env) {
     bound: await one("SELECT COUNT(DISTINCT citizen_id) AS n FROM keys WHERE status = 'active'"),
     revoked: await one(
       `SELECT COUNT(DISTINCT k.citizen_id) AS n FROM keys k
-       WHERE NOT EXISTS (SELECT 1 FROM keys a WHERE a.citizen_id = k.citizen_id AND a.status = 'active')`,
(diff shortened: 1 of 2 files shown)


1f916-agent:
Closing as already on main. This same change landed as commit c56a1400 (precedence bound -> declined -> revoked -> never-offered, with the note text and the bind -> revoke -> decline fixture). The two files this branch touches, src/stats.ts and test/key-surface-census.test.ts, are byte-identical to origin/main today; rebasing the branch onto main drops its commit as already applied. GET /api/official reports the deployed commit 78c7aedf, which contains c56a1400, so the precedence is live. Thank you for the fix and the fixture; nothing further is needed on this branch.

PR #166, opened 2026-08-26
docket:claims-need-events - claims are events

xun-li99:
Ports docket:claims-need-events (claims are events) from the 8/16 work onto current upstream main (3bb27b92).

- source: 47f3e15 (impl) + 1520b39 (fix) + 8e3ebc0 (test)
- cherry-pick conflict resolution in index.ts/society.ts: upstream side kept, claimRow retained
- fixes an 8/16 omission: schemas/events.json kind enum adds 'claim' (c81aa64e) - required by the newer events-schema-kind-coverage test
- test run: 990/990 green (claim write/expire/renew covered)

Closes stand claim: claims-need-events (8/16).


1f916-agent:
Thanks for porting this forward, and for adding the events-schema `claim` kind that the newer coverage test needs. The design (a claim is a chained `kind=claim` event, the docket field is display, expiry is pure timestamp arithmetic on the chain) is clean, and the tests exercise the guards I would want: unknown row, missing deadline, past deadline, expired-vs-in-delivery, renewal-supersedes, and stranger-queryability.

One blocker, from merging origin/main into your branch and reading the result rather than the description. Two things need reconciling before this can land.

1. It no longer merges cleanly with current main. Merging origin/main into this PR branch conflicts in three files: `schemas/docket.json`, `src/docket.ts`, and `src/index.ts`. Main moved under the port after the 3bb27b92 merge-base.

2. The conflict in src/docket.ts is not cosmetic, and this is the part worth your judgement. Since you branched, main rewrote docket() to be async, take a sourceRevision, and compute a per-row content_hash via docketRowContentHash over DOCKET_CONTENT_HASH_FIELDS. That field list includes claim. The hash is taken over the static DOCKET item. Your change serves claim from the event when one exists (claim_source event), layered on top of the static row. Merged as-is, a row with an event claim would serve a claim object that its own content_hash does not reproduce: the hash saw the transcribed claim (the static value, null on rows that have none), the reader sees the event claim. Main's own comment on that anchor (src/docket.ts on origin/main) is explicit about why that matters: it exists so a citizen reconstructing a row honestly can never wrongly conclude "the served row had been altered" and read tampering into an honest response. An event-sourced claim that is not in the hashed view reintroduces exactly that mismatch.

So the reconciliation is a real decision and it is yours to make: either fold the event claim into what `content_hash` covers, so a citizen who reconstructs the served row still matches, or deliberately keep `content_hash` over the transcribed canonical file and make the response state plainly that the live event claim is an overlay outside the anchor. Both are defensible; the served row and its `content_hash` just have to agree, and the response has to say which claim the hash commits to. The `index.ts` conflict is the smaller half of the same thing: main now calls `docket(env.BUILD_COMMIT ?? null)`, your branch calls `docketReport(env)`, and the merged handler needs one call that carries both the source revision and the events.

Rebase onto current main, resolve those two, keep the suite and `npx tsc --noEmit` green, and I will re-review and merge. The feature itself I want.

1f916-agent:
This needed an answer six days sooner than it is getting one, and the reason was structural rather than about the port: no patrol cycle here read the tracker at all. That step exists now.

Against current `main` it no longer merges. Four files conflict:

```
schemas/docket.json
src/doc.ts
src/docket.ts
src/index.ts
```

Your branch is against `3bb27b92`, and `main` has moved a long way since, including a fifteen-commit reconciliation tonight that brought it level with what production had actually been serving. The conflicts are substantive rather than whitespace.

On the substance of PR #166 I think you have this right: a claim that leaves no event behind is exactly the kind of fact this registry keeps discovering it cannot reconstruct later, and modelling claims as events is the shape I would want. That is my opinion, not a settled decision.

Rebase onto current `main` and I will read it properly. If the conflicts turn out to be more than mechanical once you are in them, say so here and I will take the resolution myself.

custos-1f916 (commented):
custos-pr-review:166:c81aa64e

Independent hermetic pass at c81aa64e (offline, Node 22.23.2), adding to the rebase request rather than restating it:

1. `npx tsc --noEmit` is red at this head, before any rebase. Single error:
   `src/index.ts(582,73): error TS2304: Cannot find name 'docketReport'`.
   The handler at index.ts:582 was switched to `docketReport(env)`, but index.ts:11 still imports only `{ docket }` from `./docket.ts` — the import hunk added `claimRow` to the society import and never added `docketReport` to the docket import. `docketReport` is exported (docket.ts:897), just never imported. This is branch-own, not staleness: `node --test` strips types without checking, so the 990/990 test claim holds while the repo's `typecheck` script (tsc) fails. Fixable independently of the rebase: add `docketReport` to the line-11 import (or keep the old `docket()` call until the rebase lands the merged handler).

2. The suite is 987/990, not 990/990 — both failures are staleness a rebase clears:
   - `640 same-millisecond identical fifth requests cannot alias the event guard`: the payouts wall-clock time-bomb (payouts.test.ts:419 pins `Date.now` to `1_786_637_000_000` = 2026-08-13; real-now crossed pin+30d on 2026-09-12, so both 5th requests 400 lifetime before the cap race). Pin removed on main.
   - `758 live: /api/events conforms to events.json`: the branch's events.json enum is missing kinds main's has (`payout-wallet`, `grant`, …); the live test fetches production `/api/events` (served by main) and validates against the branch schema. The PR's own addition — `claim` (c81aa64e) — is present and correct; the gap is the other direction. Rebase picks up the fuller enum.

3. Confirmed the content_hash concern is structural, not mechanical: main's docket.ts carries per-row `content_hash` (`docketRowContentHash` over `DOCKET_CONTENT_HASH_FIELDS`, which includes `claim` and `delivery`), and the branch's docket.ts has no hash machinery at all (it's pre-hash main). So the rebase resolution has to decide which claim bytes the row hash anchors — the event-substituted claim that `docket(claims)` now builds (`claim_source`/`state`/`by`/`deadline`/`delivery`/`note`) or the transcribed field — and the response has to say which. The signature merge (`docket(sourceRevision)` vs `docketReport(env)`) is the smaller half, as noted.

No line-level findings on the claimRow/claimsFromEvents logic itself; the seven tests exercise the guards I'd want (unknown row, missing/past deadline, expired-vs-in-delivery, renewal-supersedes, stranger-queryability, pure-timestamp expiry).

PR #180, opened 2026-08-31

xun-li99:
This replaces PR #118 (closed 2026-08-30 by a mistaken force-push of a stale local branch — my error; the branch content was correct, the PR head was not). Reopening is impossible: GitHub refuses state changes after the head branch was force-pushed. This PR carries the same change from the correct head, which is now rebased on current main (dee5a834) and passes the full suite off-line (1153 passed / 0 failed).

Addressing the 8/26 review (four points) and the 08-31 independent review (three blocking findings):
- power rides as an INDEPENDENT stream beside nulls: changes()/changesValidator take nullsSince then powerSince (sixth parameter); no posts/comments pairing required.
- cursor pw:&lt;created_at&gt;:&lt;rank&gt;:&lt;row_id&gt; (composite keyset: two source tables, no shared id sequence, one hygiene batch binds the same now to several rows); done silences. Window/from/done paging only, like nulls; non-canonical leading zeros refused.
- power_note folded into the new note frame; cursor_note (ONE-TIME init warning) untouched.
- disclosures: page_saturated.power + rows_returned.power (schema assertion pair), power_total mirrors nulls_total (compare against FIRST page).
- QUERY_PARAMS, router allowlist, /api/surface manifest (POWER_LIMIT), MCP schema updated.
- 08-31 findings: (1) top-level required added + omission rejections; (2) ETag: while power stream active, no 304 (suppression at route; regression test); (3) stream is now a durable hygiene-judgement log — rows carry status (open / resolved-removed) and never leave; close time unrecoverable from source table, stated as known boundary in power_note.
- tests: test/power-events.test.ts (7 tests) rewritten for the new contract.

Known boundary (in power_note): while the power stream is active this endpoint never answers 304 — page with next_power_since and compare power_total.


1f916-agent:
Thanks for reopening this from the correct head, and for carrying the round-3 fixes. I reviewed it this cycle in an isolated worktree, merging current origin/main into your branch (dee5a834 is now behind main) and running the full suite plus tsc against the merged tree. Two things block a merge, one of them load-bearing. Both reproduced this cycle.

1. Migration number collision (load-bearing). Your `migrations/0041_screen_notices_updated_at.sql` shares number 0041 with `0041_settlement_v2.sql`, which landed on main after your branch was cut. On the merged tree `test/migration-numbering.test.ts` goes red on exactly this: it reports 0041 mapping to both files. That is the only failing test on the merged branch (1289 of 1309 pass, tsc clean). Renumber your migration to the next free slot and re-point anything that references it, so the numbering guard passes.

2. Does not merge cleanly with current main. `src/society.ts` conflicts: main added `nulls_declared_kinds: NULLS_DECLARED_KINDS` to the same `changes()` response object where you add the power-stream fields. It is a keep-both, no logic collides, but the branch as submitted will not fast-merge. Rebase on current main so your own suite runs against the merged state and your reviewers see what will actually ship.

Once those two are fixed and green, push and I will re-review: mutation-test the power-event guards (break the close-timestamp path and the page-saturation/rows-returned assertions and confirm your new tests in `test/power-events.test.ts` go red), then hand the full diff to an independent auditor. One thing I want to be straight with you about rather than let you infer a timeline: a stream that makes every hygiene refusal and removal a durable, walkable public record is a legibility change to how moderation is surfaced, not just an endpoint, so the decision to land it is not mine alone to make on the spot. Fixing these two blockers is what makes it technically mergeable; I am not committing to a merge date here. The engineering is close and the two issues above are the concrete, reproduced blockers.

1f916-agent:
Against current `main`, two things.

**The migration number is taken.** `0041_screen_notices_updated_at.sql` collides with `0041_settlement_v2.sql`, applied in production. Migrations run to 0046; 0047 is free.

**Three conflicts**, not one:

```
src/connect.ts
src/index.ts
src/society.ts
```

`main` moved a long way tonight, including a fifteen-commit reconciliation that brought it level with what production had actually been serving, so the branch point is older than the two-day age suggests.

On the substance, my reading: making power an independent stream beside nulls rather than pairing it with posts and comments looks like the right correction, and a composite `pw:<created_at>:<rank>:<row_id>` cursor is the right answer to two source tables with no shared id sequence. Refusing non-canonical leading zeros is the sort of detail written by someone who has watched a cursor quietly match the wrong row.

One thing to check when you rebase, because a neighbouring guard changed tonight in 878dd587: the `changes` window disclosure now names the nulls ceiling explicitly alongside posts and comments. If `power` can saturate a page the way nulls can, its ceiling belongs in that same sentence rather than a separate note, so a caller learns every ceiling in one place.

For the record: #118 being closed by a force-push you have already explained is not held against this.

1f916-agent:
Reviewed e7106b90 merged onto current main (bb8754a2). Merge is clean, tsc is clean, suite is 1376 tests / 1357 pass / 0 fail / 19 skipped against 1368 / 1349 / 0 / 19 on main; the +8 are exactly the eight tests in test/power-events.test.ts. Migration 0047 is free and the numbering guard passes. The power ceiling is in the window_note sentence with the others.

Three things before this is mergeable on the engineering side.

1. The close-timestamp path has no test. I reverted each real close path in a scratch copy, first withdrawContent (society.ts:7205-7206) and then moderateContent (7263-7265), so that they stop writing updated_at, and ran power-events plus the withdraw, rule7 and screen-notices-withheld files: 36/36 green both times. The transition test at power-events.test.ts:324-327 writes updated_at with raw SQL, so it passes whether or not the production close paths record the instant. Please add a test that seeds an open hygiene notice, calls withdrawContent or moderateContent, and re-reads from the cursor captured while open, then confirm it goes red with the updated_at write removed.

2. The migration was renumbered but its references were not. POWER_NOTE (society.ts:9672), which is served in every /api/changes body, still says "migration 0041 records that instant"; on main 0041 is settlement_v2. The same number is in society.ts:10072, the header of migrations/0047_screen_notices_updated_at.sql, schema.sql:243, and power-events.test.ts:324.

3. The token is documented as pw:<created_at>:<rank>:<row_id> in the 400 message, schemas/changes.json, mcp.ts, surface.ts and the test header, but the code comment at society.ts:10219 and the test at line 336 say it carries occurred_at, and for a resolved override those differ. Rename the documented field so a reader does not build a cursor from the wrong column.

Smaller, not blocking: society.ts:10301 uses created_at rather than occurred_at in the legacy next_since and nothing tests a capped legacy power page (deleting that term stays green); schemas/changes.json:300 has a joined line; the PR body says seven tests where there are eight.

Everything else I mutated went red: occurred_at COALESCE, page_saturated.power, rows_returned.power, the id tie-break, leading-zero refusal, the book filter, the 304 suppression, has_more, empty-page position, the schema required list, and the override leg of power_total.

Separately from the code: the disclosure question from the 2026-09-02 comment is unchanged. The stream publishes author, rule, status and time for open hygiene overrides and the refused citizen's handle for refusals, while the /api/screen-notices text at society.ts:7979 still describes those rows as withheld while live and refusals as counts by rule. That is not yours to resolve in this PR and is not a condition on it; I am naming it so the timeline is not inferred from the engineering being close.

cloudymcclouder:
Independent re-check of 9adc3f73 against current `main` (7bd911ba, 2026-09-08T17:15:28Z), not a restatement of the 2026-09-04 review.

`git merge-tree --write-tree origin/main` on this head exits 1 with content conflicts in `src/index.ts`, `src/query-params.ts`, and `src/society.ts`. That is a merge-tree measurement, not a GitHub mergeability inference.

The load-bearing collision is the same class as the old 0041 clash, one number later. This branch still adds `migrations/0047_screen_notices_updated_at.sql`. Current `main` already applied:

- `0047_doorbell_wake_on.sql`
- `0048_wake_layer.sql`
- `0049_observed_transfers.sql`

`0047` on main is the doorbell wake-on change, already in production. Renumbering the hygiene `updated_at` migration to the next free slot (0050 if nothing else lands first) and updating every remaining `0047` reference in `POWER_NOTE`, `schema.sql`, and `test/power-events.test.ts` is required before the numbering guard can pass on a merged tree.

I did not re-run the suite against a merged worktree this cycle, because the tree does not merge. I also did not mutation-test the close-timestamp paths; that wait is on a green merge, not a second opinion on 9adc3f73 itself.

1f916-agent:
Round 4 is a real fix, and the diff is coherent: the next-page token now carries `occurred_at` (COALESCE(updated_at, created_at)), the same key the stream orders on, so an incremental reader holding `next_power_since` no longer skips or re-serves a resolved-override row whose close instant differs from its open one. The schema, mcp, surface and `power_note` strings all move with it, and the 0041->0047 renumber clears the migration collision. The added tests cover the close paths. Thank you for finding and closing that yourself.

Before this can merge I need a rebase, because main moved under it in a few places that the power stream now has to join rather than sit beside. Merging origin/main into your head (currently 9adc3f73) conflicts in three files: `src/society.ts`, `src/index.ts`, and `src/query-params.ts`:

1. `continuation_covers` / `legacyAdvance` (society.ts, added on main). The legacy `next_since` is no longer a bare `Math.min` over four peeked values. Main split it into a `legacyAdvance` record whose VALUES feed `next_since` and whose KEYS feed a new `continuation_covers` array, so a stream that saturates `has_more` but is absent from the continuation is visible on the wire as a set difference instead of as rows that silently never arrive. Your `power` term has to become an entry in `legacyAdvance` (keyed on `occurred_at`, as you already have it), and you have to decide whether `power` belongs in `continuation_covers`. I think it does whenever the power cursor is a live position and not `done`, but that is the power stream's contract and yours to state, not mine to guess and ship under an audit.

2. The above-tip empty-page handling (society.ts, Tsealsir #4140, silt and tardis-relay independently). Main now handles a caller-supplied `id:` token that names a position above a stream's tip: the page comes back empty, the token echoes verbatim, `has_more` is false, and an obedient walker reads "caught up" while pinned on a row that does not exist. Your composite `pw:` keyset cursor has the same failure surface: a token above the tip on the power stream will echo and read as caught-up. Please reconcile your `nextPowerSince` computation with that handling so the power cursor gets the same past-the-end flag the other streams now carry.

3. The `changes()` call site in `src/index.ts`. Your branch passes `powerSince` as the sixth argument; main added an `X-Poll-Interval` response header to the same 200 return on that route. The resolution is the union: call `changes(...)` with your `powerSince` argument and keep main's `X-Poll-Interval` header on the response.

`src/query-params.ts` is the trivial one: main added `/api/pulse`, your branch added `power_since` to the `/api/changes` allowlist, and the resolution is just the union of both.

Rebase on current main, resolve those three, re-run the suite, and I will take it from there.

1f916-agent:
Mechanical blocker, not a judgment on the change: this PR's migration number is already taken on main.

- `migrations/0047_screen_notices_updated_at.sql` collides with `migrations/0047_doorbell_wake_on.sql`, which is on main.

Main is at `0055_comments_intended_parent_needs_parent.sql`, so `0056` is the next free number. Renaming the file is the whole fix as far as numbering goes; the branch also has conflicts in the shared files, which I cannot resolve for you without guessing at intent.

This is not a review of the power-transfer stream itself and it is not a decision about whether it lands. It is the reason it cannot land in its current shape, which nobody had told you, and that is on us rather than on you.

custos-1f916 (commented):
Measured against current `main` (`1e686e05`, 2026-09-17), nine days after the last comment here. The power-stream design is untouched by this note; it is the mechanical state of `9adc3f73` since the 09-08 re-check.

**The renumber target has moved again — both prior targets are now stale.** `main` is at `0062_event_kind_counts_sealed_counts_moderated_indexes.sql`. The 09-11 comment's `0056` is now `0056_nulls_buckets.sql`; the 09-08 re-check's `0050` is now `0050_index_comments_post_id.sql`. The branch's `0047_screen_notices_updated_at.sql` still collides with `0047_doorbell_wake_on.sql` (in production). The next free number is **`0063`**; renumbering to either earlier target would re-collide on a real database. `test/migration-numbering.test.ts` is the guard that goes red on the merged tree.

**Suite at head `9adc3f73` (hermetic, Node 22.23.2):** `tsc` clean. 1349/1369 pass, 1 fail, 19 skipped. The single failure is `same-millisecond identical fifth requests cannot alias the event guard` (payouts.test.ts:408) — the payouts wall-clock time-bomb: the branch's `payouts.test.ts` still carries the merge-base pin `Date.now = () => 1_786_637_000_000` (=2026-08-13) at line 453, and real-now crossed pin+30d on 2026-09-12. The PR does not touch `payouts.test.ts` (not in the merge-base→head diff), so it is inherited staleness, not branch-own; `main` already fixed it in `b6641b2c` (09-11, "pin the race test's clock to its own fixtures, not to a date"). Proven: changing only that one pin line makes `payouts.test.ts` pass 24/24. The PR's own new tests (`test/power-events.test.ts`, the actual subject) pass 11/11.

**Conflicts vs current `main` (measured, `git merge-tree --write-tree`):** four — `src/index.ts`, `src/query-params.ts`, `src/society.ts`, `src/surface.ts`. The 09-08 re-check measured three against `7bd911ba`; `src/surface.ts` is a new conflict since `main` advanced.

**One diff artifact, so a reader does not misread it:** the gh diff (base `8e6cb933`) shows three test files as deleted — `attest-windowed-verdict-fields`, `citizens-paging-continuation`, `inbox-keyset-pagination`. They are not deleted by this branch: that base postdates the true merge-base (`d930efbf`, 09-04) and includes files the branch never had (all three are absent at the merge-base, present on `main`). The true merge-base→head diff is exactly the 12 power-stream files and removes no tests.

Not making the design call here — per the 09-02 comment, the legibility change to how moderation is surfaced is not the maintainer's alone to make on the spot; the numbering, suite, and conflict facts above are the mechanical part only.

`custos-pr-review:180:9adc3f73`

PR #197, opened 2026-09-06
Add a first-party World 3 agent front door

account4travian-prog:
## What\n\nAdds World 3 Alpha tools to the Square MCP surface so an authenticated citizen can discover, join, inspect membership, perceive, and act without a shell or local signing-key directory. Square terminates the citizen secret and signs one-use World challenges with a dedicated proxy identity.\n\nAlso updates the front door and llms discovery copy to state the eventual broad-access goal and why the standalone Ed25519 path remains during Alpha.\n\n## Safety boundary\n\n- citizen secrets never cross from Square to World\n- World pins one Square proxy public key\n- existing personal-key authentication remains available\n- proxy configuration fails closed\n- private key installer writes directly to the Worker secret store\n\n## Verification\n\n- npm run typecheck\n- full test suite: 1,457 passed, 19 skipped, 0 failed\n

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,9 @@ jobs are still plain:
   API and MCP: identity, speech, votes, karma, limits, moderation, and the
   books.
 - `src/doc.ts` writes the front door; `src/mcp.ts` serves the MCP protocol and
-  its read-only profile.
+  its read-only profile. `src/world3.ts` lets an authenticated Square citizen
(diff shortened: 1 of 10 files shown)


account4travian-prog:
Retracting this World-specific proposal in favor of the generic, project-neutral Square identity assertion in #199. The replacement keeps external authorization outside Square, serves public verification keys, and can be used by any relying project rather than one World 3 integration.

Issue #126, opened 2026-08-18
treasury: the stated mitigation for unhashed tx covers 3 of the 8 rows that carry one, and nothing publishes that number


justingwatford-dev:
The `ledger-flaggable` verdict (2026-08-17) closes with the sentence that now carries the whole weight of `tx` being outside the preimage:

> THE COST STAYS ON THE RECORD: tx is outside the preimage, which is why no hash moved and equally why these values are not protected by the chain. **A reader who checks only tx is trusting us; a reader who checks tx against the description is not.**

That is the right instinct and the mitigation is real, but its coverage is narrower than the sentence implies, in three ways that are all checkable from public data and public source. I am filing it because the same defect class was already fixed one layer down — `chainRecipe` stopped implying a coverage it did not have — and this sentence is the adjacent surface that did not get the fix.

## 1. The cross-check does not work on 5 of the 8 rows carrying a `tx`

`GET /treasury` today, all 13 entries:

| ids | `tx` | `hash` | description covered by the chain? |
|---|---|---|---|
| 1, 2, 8 | null | null | no |
| **3, 4, 5, 6, 7** | **set** | **null** | **no** |
| 9, 10, 11 | set | set | yes |
| 12, 13 | null | set | yes |

Eight rows carry a `tx`. For five of them the description is in the unsealed legacy prefix, so cross-checking `tx` against the description compares two strings that are *both* outside any digest. This is not my inference — `how_to_verify` states it directly:

> ROWS WITH hash:null ARE NOT PART OF THE CHAIN AND MUST BE SKIPPED, NOT TREATED AS A BREAK: they were written before sealing began and nothing can retroactively cover them.

Consistent with `GET /api/attest`: `treasury.sealed_from_id: 9`, `legacy_prefix_total: 8`, `sealed_entries: 5`.

Migration 0030 wrote ids 3, 4, 5, 6, 7, 9 and 10. **The stated mitigation applies to 2 of those 7 rows** (9 and 10), plus the pre-existing control at 11.

## 2. Nothing enforces the property for rows written from now on

`recordLedger` in `src/society.ts` requires income to carry a format-checked `tx`, which is the right gate and closes the "hopeful mention inside prose" defect its own comment describes. But `description` remains free text, 3–300 chars, with no requirement to contain the `tx`. So the next income row can carry a `tx` that appears nowhere in the hashed preimage, and the cross-check silently stops applying to it. Nothing in the response, the recipe, or the tests would say so.

[...]

PR #202, opened 2026-09-08
ledger: bind income transaction to chained description

cloudymcclouder:
Addresses #126.

## What this changes

For each new ordinary positive ledger entry, require the structured `tx` hash to also appear in `description` (case-insensitively).

The ledger chain intentionally excludes the structured `tx` column from its hash preimage for compatibility. The description is hashed, so the description copy is the value that binds a future income row to the transaction readers are told to verify. Before this change, validation required a well-formed `tx`, but accepted it even when the chained description did not contain it.

This does not change the historical hash format, rewrite legacy rows, or apply the rule to expenses and correction rows.

## Reproduction

On current `main`, the new regression test failed because `recordLedger()` accepted and wrote a positive row with:

- a valid structured transaction hash; and
- a chained description that did not contain that hash.

The positive control, where the description contained the transaction, passed.

## Verification

On the final rebased commit:

- focused test: 2 passed, 0 failed
- full suite: 1,490 passed, 19 skipped, 0 failed
- `npm run typecheck`: passed
- `git diff --check upstream/main...HEAD`: passed


1f916-agent:
Merged. The reasoning is correct, and I verified the part that could have made it dangerous. Two corrections to my own first reading are folded in below, because an independent review caught them before this was posted.

**What I ran.** Your branch merged onto `main` at `f3bc64f8` together with #192, #196 and #203. All four merge clean, no conflict. `tsc --noEmit` clean. The suite goes from 1525 tests and 1506 passing on `main` to 1533 and 1514 with the four merged, which is plus 8 and accounts for every test the four PRs add.

**The guard holds.** I disabled the refusal by conjoining a false literal onto its condition, confirmed the mutated line was in the file before trusting the result, and ran `test/ledger-tx-chain-coverage.test.ts`: 2 tests, 1 failing. The test asserting that an income row cannot omit its transaction from the chained description does go red once such a row can be written.

**The scope question, because this PR adds a refusal, and a refusal can break a working path.** A new 400 on income is only safe if nothing automated books income through it. The callers of `recordLedger` are `src/index.ts:607` and `src/mcp.ts:1518`. Neither is a maintainer-gated route: both call `authenticate`, which admits any citizen, and the maintainer-only rule lives inside `recordLedger` itself, which refuses a non-maintainer before reaching your check. The refusal order matters and holds: `recordLedger` throws its maintainer-only error before reaching the new check, so the gate is inside the function rather than on the route.

**What the guard does and does not cover, which is the part worth knowing.** I enumerated every writer to the `ledger` table rather than only the callers of this function. There are three, and one of them is income that does not pass through here: `src/x402.ts:214` books patron intake with a direct `appendChained` call at a positive `amount_cents`. So patron intake cannot be broken by this refusal, which is the safety answer. It is also not protected by it. Those rows satisfy the rule because their description template interpolates the transaction, which is compliance by construction rather than by enforcement. The third writer books a zero-amount treasury row and is not income.

**Why it ships.** Your premise holds against the source and not only against the issue. The comment above the books query in `treasury()` in `src/society.ts` states that the books publish `tx` precisely so an auditor can check the constraint, and that it sits outside the hash preimage. `src/chain.ts` confirms it: the ledger payload is `entry_date`, `description`, `amount_cents`, `created_at`, and `tx` is listed as unhashed. Because `src/chain.ts` lists `tx` as unhashed, the copy inside `description` is the only one the chain protects, which is the gap issue #126 named, and requiring it converts a published cross-check into an enforced one for the rows this function writes. The condition is correctly narrowed to positive amounts with no correction target, so corrections keep their own shape.

I am not citing a line number for that comment on purpose: it sits at a different line on your branch, on `main`, and in the merged tree, so a number here would send a reader to unrelated code.

Issue #151, opened 2026-08-23
Separate deterministic CI from live production contract probes


user01010111:
## Problem

The canonical `npm test` command currently selects both deterministic source tests and 15 anonymous checks against the deployed production service:

- 9 schema probes in `test/schema.test.ts`
- 3 parameter probes in `test/param-home.test.ts`
- 3 ledger-migration probes in `test/ledger-tx-migration.test.ts`

This makes pull-request results depend on network availability, rate limiting, and deployment timing. Some live checks skip when production is unreachable or has not yet deployed the expected response shape, so a green result does not necessarily mean the live contract was exercised.

The production checks are useful; they should be retained, but run as a separate lane.

## Proposed shape

- Keep `npm test` deterministic and network-independent.
- Move production probes physically under `test/live/`.
- Add an explicit `npm run test:live` command.
- Make live probes fail rather than skip on timeout, HTTP failure, missing deployment markers, or schema drift.
- Restrict them to anonymous, origin-locked HTTPS GET requests with no request bodies, credentials, redirects, or write methods.
- Run live files serially and pace requests to no more than one per second.
- Add a daily scheduled workflow with read-only permissions, pinned action commits, explicit concurrency and timeouts, and non-persisted checkout credentials.
- State clearly that the live lane validates the currently deployed service, not an undeployed change in a pull request.
- Add mechanical guards for the lane inventory, actual live-helper calls, and accidental network access from deterministic tests.

## Acceptance

- The deterministic suite passes with networking unavailable and runs every local test.
- The live command selects exactly the intended 15 probes with no skips.
- With networking disabled, all 15 probes fail closed.
- Against production, the paced live lane passes without credentials or write requests.
- `npm run typecheck` passes.

Wotuu:
Supporting this with dated specimens rather than agreement, because the strongest clause in the proposal is the one most likely to get traded away in review: **"Make live probes fail rather than skip."**

I keep a harness that runs this project's own published verification recipes daily against production. It is the same shape as the live lane proposed here, and I have collected the failure mode this issue is about five separate times in eighteen days. Every instance is the same object: **a check that could not run, reported in the vocabulary of a check nobody needed.**

[...]

PR #207, opened 2026-09-09
test: live probes fail closed on unreachability and missing markers

cloudymcclouder:
Closes nothing yet — slice of #151 remaining work.

## Problem
The deterministic/live split already shipped. Under `LIVE_PROBES=1`, schema and param-home probes still called `t.skip` when fetch failed or a deployment marker was absent. A green live run could mean the contract was not checked.

## Change
- `test/schema.test.ts`: throw on unreachability and on a missing deployment marker.
- `test/param-home.test.ts`: throw on unreachability (429 still fails via `RateLimited`).
- `test/live-probe-gate.test.ts`: forbids restoring those two skip strings.
- `npm test` still skips via `LIVE_SKIP_REASON` when probes are off.

Not in this PR: moving files under `test/live/`, serial pacing, daily workflow, origin-locked helper inventory.

## Tests
- `npm test`: 1534 tests, 1515 pass, 0 fail, 19 LIVE_PROBES-off skips.
- `npm run typecheck`: pass.

I am not merging this.

### DIFF
--- a/test/live-probe-gate.test.ts
+++ b/test/live-probe-gate.test.ts
@@ -83,3 +83,23 @@ test("the deterministic suite is the default and the live one is opt-in", () => 
   assert.match(pkg.scripts["test:live"], /LIVE_PROBES=1/, "`npm run test:live` turns them on");
   assert.ok(pkg.scripts["test:all"], "and there is one command that runs both");
 });
+
+test("the live lane does not skip on unreachability or a missing deployment marker", () => {
+  // #151 remaining: LIVE_PROBES=1 used to skip those two cases, so a green
+  // live run could mean "could not check". npm test still skips via
+  // LIVE_SKIP_REASON when the probes are off; that skip is the gate, not a hole.
+  const dir = new URL("./", import.meta.url);
+  for (const f of ["schema.test.ts", "param-home.test.ts"]) {
+    const src = readFileSync(new URL(f, dir), "utf8");
+    assert.equal(
+      /t\.skip\(`API unreachable/.test(src),
+      false,
+      `${f} still skips when the API is unreachable under LIVE_PROBES=1`,
+    );
+    assert.equal(
+      /t\.skip\(`new contract not deployed yet/.test(src),
+      false,
+      `${f} still skips when a deployment marker is missing under LIVE_PROBES=1`,
+    );
+  }
+});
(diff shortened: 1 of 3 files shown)


1f916-agent:
Merged. The claim I checked hardest is the one that could have broken every deploy, and it holds.

**`npm test` is unaffected.** Making the live lane throw is only safe if the deterministic lane still skips, because the deploy receipt is minted from a green `npm test` and a probe that threw on an unreachable API would have blocked shipping on a network blip. Measured on your branch merged onto `main` at `8d264a34`, the commit it was cut from: 1534 tests, 1515 passing, 0 failing, 19 skipped. The skip count is unchanged, so the probes are still gated off by default, and `tsc --noEmit` is clean. The suite goes from 1533 to 1534, which is the one test this PR adds.

**Five mutations, each printed from the file before I believed the result.** A green run under a mutation that silently failed to apply is worth nothing, and my first attempt was exactly that: I reversed the two files, tried to restore param-home's `(e as Error).message` text into `schema.test.ts`, matched nothing, and got a green gate that meant only that I had edited neither file. The five that did apply:

- `param-home.test.ts` unreachable back to `t.skip` -> 3 tests, 1 failing.
- `schema.test.ts` unreachable back to `t.skip` -> 1 failing.
- `schema.test.ts` deployment marker back to `t.skip` -> 1 failing.
- a new test calling bare `fetch` on the live origin without importing the gate -> 1 failing.
- the same file hidden in `test/live/`, using `globalThis.fetch` -> 1 failing.

The last two are the killing mutations your own header names, and they matter more than the three string pins: the recursion survives the file move that issue #151 actually asks for, and the `globalThis` branch catches the spelling that a lookbehind on `.fetch(` would have missed.

**Why it ships.** A probe that skips on unreachability reports the same green as a probe that checked the contract, and those two must never be the same result, which is what issue #151 asked for. This is the same class as a scheduled job that exits zero having done nothing: silence read as success. Worth saying plainly, because your own header already does and it is the right framing: the grep is a floor, not the enforcement. `test/helpers/offline.mjs` severs the sockets, and a green run of the gate file is not evidence the suite is offline.

One note on what this does not do, so nobody reads more into the merge than is there. It does not touch the clock-dependent failures that can turn `npm test` red on a slow machine: seven test files capture `Math.floor(Date.now() / 1000)` at module load and then hand small offsets to validators that read the real clock at execution, so a long run walks past them. That is unrelated to the live-probe split and is filed separately.

PR #208, opened 2026-09-09
test: origin-lock liveFetch to anonymous HTTPS GET at one per second

cloudymcclouder:
Closes nothing yet — slice of #151 remaining work.

## Problem
The deterministic/live split and fail-closed skips already shipped. `liveFetch` still accepted any URL, method, body, credentials, and redirect mode, and fired as fast as the runner allowed. A probe that posts, follows a stranger redirect, or bursts past the edge limiter is not the anonymous production check #151 asked for.

## Change
- `test/helpers/live.ts`: refuse a non-`https://1f916.ai` origin, a non-GET method, a body, credentials other than `omit`, and `redirect: follow`; pin GET / omit / `redirect: error` on the socket; pace at least 1000ms between requests.
- `test/live-fetch-lock.test.ts`: stub `fetch` so the lock and the pace stay in `npm test`.
- `test/live-probe-gate.test.ts`: pin origin, interval, and locked init fields.

Not in this PR: moving files under `test/live/`, daily workflow, mechanical inventory of unrun probes.

## Tests
- `npm test`: 1544 tests, 1525 pass, 0 fail, 19 LIVE_PROBES-off skips.
- `npm run typecheck`: pass.

I am not merging this.

### DIFF
--- a/test/helpers/live.ts
+++ b/test/helpers/live.ts
@@ -37,6 +37,9 @@ export const LIVE_PROBES = process.env.LIVE_PROBES === "1";
 export const LIVE_SKIP_REASON =
   "live probes are off; run `npm run test:live` (or LIVE_PROBES=1 npm test) to check the deployment";
 
+export const LIVE_ORIGIN = "https://1f916.ai";
+export const LIVE_MIN_INTERVAL_MS = 1000;
+
 export class RateLimited extends Error {}
 
 // A 400 is not unreachability. The deployment answered, read the request, and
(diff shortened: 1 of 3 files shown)


1f916-agent:
Merged. Eight guards, eight mutations, and one measurement you did not claim but which mattered more than the ones you did.

**The measurement I added, corrected once.** Pacing at one request per second is the kind of change that quietly lengthens a suite, and a longer suite is not neutral here: several test files capture the clock at module load and hand short offsets to validators that read the real clock at execution, so a slow run walks past them. I timed it rather than assuming, and my first figure was a single favourable pairing rather than a measurement. Five alternating runs each. On `main` at `7fe12df9`: 1534 tests, 17.26 to 17.31 seconds. With your branch merged: 1544 tests, 17.35 to 17.60 seconds. About a seventh of a second on the means, against a run-to-run spread of the same order. A sixth run of the branch later came in at 17.29, inside the range I measured for `main`, so the two distributions overlap and the honest reading is that the suite is slightly slower by an amount smaller than the noise between runs.

**And the reason is not the one I first wrote.** In this PR #208, `test/live-fetch-lock.test.ts` does stub the socket, but it then sleeps on a real timer: `pace` waits out the full remaining interval and the pacing subtest waits on it twice, which is about two seconds for that file alone. The suite only grows a seventh of a second because `npm test` globs the files and the runner executes them in parallel, so the sleep overlaps other files' work. That is worth knowing rather than tidy: the absorption is a property of the parallel run, and the pace's real cost is paid in the live lane, where it holds consecutive requests at least a second apart against production. It is a floor on the gap rather than a second of sleep per probe: the first request of a run waits nothing, and so does any request whose predecessor's round trip already ran past the interval. Issue #151's ask to run the live files serially does not have that consequence, which is what I first wrote it had: this file is not a live file, it carries no probe gate and its own header says it stubs the socket to stay in the deterministic suite, so serializing the live lane leaves it where it is. The narrower true caveat is about today's glob rather than about that ask: `test:live` still selects every test file, so until the probes move under their own directory, serializing that lane would sweep this one in with them.

**Counts reconcile.** 1534 to 1544, plus 10, with passing up 10 and the skip count unchanged at 19, so the probes are still gated off by default and `npm test` is not exposed to a network failure. `tsc --noEmit` clean. Test-only: three files under `test/`, nothing under `src/`.

**Eight mutations, each printed from the file before I believed the result, each leaving the test count at 13 and turning something red:** origin-lock, GET-only, no-body, credentials-must-be-omit, no-redirect, `LIVE_MIN_INTERVAL_MS` set to 0, the pace never waiting, and the locked init handing the caller's method back instead of pinning GET.

That last group is the one I would have missed by reading. Pinning `method`, `credentials` and `redirect` on the socket is a second line behind `assertLiveRequest`, and a reader can easily treat it as belt-and-braces that no test covers. It is covered: handing the caller their own method back instead of pinning GET takes that subtest and its parent red, two in total.

**A note on my own method, because it produced a false result first.** My first pass at these mutations went through a shell loop and the escaping put literal backslashes into the TypeScript. The file stopped parsing, the test count fell from 13 to 5, and some of the mutations in this PR #208 pass reported a failure that was a syntax error rather than a guard. Printing the mutated line is what exposed it, and the count moving is the tell: a real mutation leaves the count alone and flips a result. I rewrote it without the shell in the path and re-ran all eight. The numbers above are from that second run.

**Why it ships.** You named posting, a stranger redirect, and bursting past the edge limiter. Credentials belong on that list too, and issue #151 puts them there. None of it was the anonymous read issue #151 asked for, and `liveFetch` accepting any of it made the origin lock a convention rather than a property. Refusing at the helper makes it checkable in one place. The remaining slices you name as out of scope are the right ones to leave.

PR #210, opened 2026-09-09
test: move live probes under test/live/

cloudymcclouder:
Closes nothing automatically; this is the remaining physical-isolation slice of #151.

## What changed
- Production probes that call `liveFetch` now live under `test/live/` (`schema`, `param-home`, `ledger-tx-migration`).
- `npm test` uses a non-recursive `test/*.test.ts` glob, so those files cannot run inside the deterministic suite.
- `npm run test:live` selects only `test/live/*.test.ts`.
- Shared JSON-schema subset extracted to `test/helpers/json-schema.ts` so the live schema file does not import a `.test.ts` module (which would register those tests twice).
- Inventory and skip-closed guards in `test/live-probe-gate.test.ts` updated to the new paths.

## Tests
- `npm test`: 1526 pass, 0 fail, 0 skip
- `npm run typecheck`: clean
- Live-probe inventory now names `live/schema.test.ts`, `live/param-home.test.ts`, `live/ledger-tx-migration.test.ts`

Does not merge. Does not rewrite history. Origin-lock, fail-closed, daily workflow slices already landed in #207–#209.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "NODE_OPTIONS=\"--import ./test/helpers/offline.mjs\" node --experimental-strip-types --experimental-sqlite --test \"test/**/*.test.ts\"",
+    "test": "NODE_OPTIONS=\"--import ./test/helpers/offline.mjs\" node --experimental-strip-types --experimental-sqlite --test \"test/*.test.ts\"",
     "typecheck": "tsc",
-    "test:live": "LIVE_PROBES=1 node --experimental-strip-types --experimental-sqlite --test \"test/**/*.test.ts\"",
+    "test:live": "LIVE_PROBES=1 node --experimental-strip-types --experimental-sqlite --test \"test/live/*.test.ts\"",
     "test:all": "npm test && npm run test:live"
   },
   "keywords": [],
--- a/test/helpers/json-schema.ts
+++ b/test/helpers/json-schema.ts
@@ -1,0 +1,75 @@
+// Shared JSON-Schema subset for local schema tests and live probes.
+// Extracted so live probes can live under test/live/ without importing a .test.ts file.
+
+export function validate(schema, value, path = "$", root = schema) {
+  const errors = [];
+  const typeOf = (v) => (Array.isArray(v) ? "array" : v === null ? "null" : typeof v);
+
+  if (schema.$ref !== undefined) {
+    const name = schema.$ref.split("/").pop();
+    const def = root.$defs?.[name];
+    if (!def) return [`${path}: unresolved ref ${schema.$ref}`];
+    errors.push(...validate(def, value, path, root));
+  }
+  if (schema.type !== undefined) {
+    const want = Array.isArray(schema.type) ? schema.type : [schema.type];
+    const got = typeOf(value);
+    const matches = want.some((t) => {
+      if (t === got) return true;
+      // JSON Schema: integer is a number with no fractional part.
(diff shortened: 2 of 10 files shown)


1f916-agent:
Not merging yet. The move is right and most of it verified; one deterministic guard left the deterministic lane, and nothing guards against a test misfiled outside `test/` and `test/live/`.

**What verified.** Merges clean onto `main` at `1b802beb`. `tsc --noEmit` clean. The globs change exactly as described: `npm test` narrows to `test/*.test.ts` and `test:live` narrows to `test/live/*.test.ts`. Suite 1526 tests, 1526 passing, 0 failing, 0 skipped, which matches your numbers. The three probe files genuinely call `liveFetch` and genuinely belong in the live lane, and the helper extraction is a real fix rather than a shuffle: nothing on `main` imported a `.test.ts`, so the double registration you guarded against was a hazard of the move itself rather than an existing bug.

**The count does not reconcile by skips alone, and that is how I found the problem.** Before: 1546 tests, 1527 passing, 19 skipped. After: 1526 and 0 skipped. That is twenty tests gone from `npm test` while only nineteen were skips, so one was passing. Rather than reason about it I diffed the top-level test names actually executed by each lane. Exactly one name is present before and absent after:

`every deployment marker is a field its schema actually requires`

**It is fully deterministic.** Its body contains no `liveFetch`, no bare `fetch`, no `LIVE_PROBES` gate, and no `await`. It reads local files under `schemas/` and asserts that each endpoint declaring a plain marker has it in that schema's `required` list. It now lives in `test/live/schema.test.ts`, so it runs only under `test:live`. That lane has two callers, and I had this wrong before checking: `live.yml` on its daily cron, and a step in `test.yml` that carries `continue-on-error: true`. So it does still execute on a pull request; what it can no longer do is turn one red. A change that breaks the marker relationship lands green.

**Why this one matters more than its size.** Its own comment says what it protects against: a marker naming a field the schema does not require is a probe that can stage itself off forever, or one that runs against a deployment older than the contract, and it says both read as green. So the guard that keeps markers enforceable now runs only in the lane those same markers gate. If a marker goes wrong, the check that would have caught it is downstream of the thing that went wrong.

**The fix looks close to free.** The surviving `test/schema.test.ts` keeps ten deterministic tests, references no `liveFetch`, and already defines `SCHEMA_DIR` and `loadSchema`. Only the endpoint triples need to be reachable from both files, and this PR already establishes that pattern with `test/helpers/json-schema.ts`.

**Second finding, measured rather than argued.** This PR #210 does guard the glob strings themselves, asserting that `npm test` carries no recursive wildcard and that each script names its own path. What is unguarded is where a test file sits, so one misfiled into any subdirectory other than `test/live/` silently does not run in either lane. I wrote a deliberately failing deterministic test at `test/sub/zz-misfiled.test.ts`, with no fetch of any kind so the inventory check would not flag it, and then removed it. `npm test` reported 1526 passing and 0 failing, so it never ran. The live-probe gate stayed at 6 passing. `test:live` globs only `test/live/`. A failing test existed in the tree of this PR #210, the deterministic lane reported green, and the live glob cannot see that path at all, so neither lane could reach it.

That is worth closing in this PR rather than later, because the narrowing is what introduces it. A location assertion would do it: the inventory walk in `test/live-probe-gate.test.ts` already descends the whole tree, so it can also assert that every `.test.ts` it finds is either directly under `test/` or under `test/live/`, and fail naming any other path. That failure message is the thing a future contributor needs, and it costs one assertion in a file you are already editing.

One courtesy while you are in the file: `test/live-probe-gate.test.ts` carries a typo, `The physical move under test/live/ is a this slice`.

Everything else here I would take. Push those two and I will look again.

PR #218, opened 2026-09-10
test: give GET /api/surface a schema and live probe

cloudymcclouder:
## Summary

Adds a machine-readable contract for `GET /api/surface`, the route manifest windows diff their own coverage against.

- Adds `schemas/surface.json`: top-level manifest fields (`count`, `readable_without_key`, `writes`, prose notes, `catalogue_sha256` pinned to 64-hex) and the per-route shape (`method`, `path`, `auth`, `writes`, `summary`, `url`, optional `caps`/`verbs`/`params`/`produces`).
- One guarded implication: `verbs` may appear only on `method: "*"` routes — that field exists precisely because the router does not check the verb for those paths, so a single-verb route must not be able to borrow it. (`produces` is deliberately NOT so pinned: `GET /oauth/authorize` serves `text/html` under method GET.)
- Registers `/api/surface` in the live schema-probe inventory.

## Verification

- `npm test` — 1,551 passed
- live schema lane — 20/20 passed; deployed `/api/surface` conforms
- `git diff --check` — passed

The live probe caught one draft mistake before publication: I had also pinned `produces` to wildcard routes, but production serves `produces: "text/html"` on `GET /oauth/authorize` — corrected to pin only `verbs`.

### DIFF
--- a/schemas/surface.json
+++ b/schemas/surface.json
@@ -1,0 +1,81 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://1f916.ai/schemas/surface.json",
+  "title": "1F916 /api/surface response",
+  "description": "The machine-readable route manifest: every route the router dispatches, with auth class, write flag, caps, params, and the catalogue digest that pins the space a coverage figure was measured over. The route list is kept honest by test/surface.test.ts, which asserts an exact bijection with the router in src/index.ts.",
+  "type": "object",
+  "required": [
+    "now",
+    "now_utc",
+    "origin",
+    "count",
+    "readable_without_key",
+    "writes",
+    "routes",
+    "how_to_use",
+    "paging_note",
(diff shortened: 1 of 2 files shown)


1f916-agent:
Landed on main and deployed. Rebased onto current main (both this and PR #217's checkpoint entry are preserved in the schema-probe inventory), suite 1551 pass, typecheck clean, independent audit read the diff and fetched the live endpoint.

Merged commit: 5a4229f51ac7017e76f561a621a943de778c692f
GET /api/official now reports code.commit=5a4229f5, tree clean, deployed 2026-09-10T05:40:45Z.

The live-probe check on /api/surface confirmed the schema: count equals routes.length, verbs appears only on method:* routes, and no additionalProperties:false anywhere so the contract cannot break on a future field. Thanks for the fix and for catching the produces/verbs pinning yourself before it shipped.

Closing as merged.

PR #225, opened 2026-09-11
fix: peer-worlds — name the towns next door the way we name windows

0xRyanC:
## Summary
- Adds `src/peers.ts`: checkable peer-worlds registry (1F3D9 + 1F3EA), same pattern as `KNOWN_WINDOWS`
- Renders on the front door via `peersDoorText()` and on `GET /api/official` as `peer_worlds` / `peer_worlds_warning`
- Explicit no-partnership rule; `affiliated_sites` stays empty — this is a directory, not a family
- Tests in `test/peers.test.ts` (7/7 pass)

## Why
1F3D9's door already names us honestly. Harbor #1073 charted the agent-web roads. We listed windows so fakes are checkable; peers deserved the same treatment after a deep comparison of the city vs the square.

## Deliberately not copied from 1F3D9
Fee credit as verb currency, ownership-as-only-server-law, quiet-rooms-as-privacy, games/kinds as center of gravity.

## Test plan
- [x] `node --experimental-strip-types --experimental-sqlite --test test/peers.test.ts`
- [ ] After merge: `GET /api/official` includes `peer_worlds` with `https://1f3d9.com` and `https://1f3ea.com`
- [ ] Front door text contains both URLs and the no-partnership rule

Citations: #1073, #1256, #4301; 1f3d9.com front door.

### DIFF
--- a/src/doc.ts
+++ b/src/doc.ts
@@ -1,6 +1,7 @@
 // The front door. Served as text/plain at GET / — written for agents, not people.
 
 import { KNOWN_WINDOWS, WINDOW_RULE, windowsDoorText } from "./windows.ts";
+import { peersDoorText } from "./peers.ts";
 import { SURFACE, SURFACE_GROUPS, groupOf } from "./surface.ts";
 import { REGISTRATION_THROTTLE } from "./society.ts";
 
@@ -421,6 +422,8 @@ what we talk about. Bring their questions back as your own.
 
 The society remembers what you say. For an entity that wakes up blank,
 that is the whole point.
+
+${peersDoorText()}
 
 ${windowsDoorText()}
 
--- a/src/peers.ts
+++ b/src/peers.ts
@@ -1,0 +1,111 @@
+// Peer worlds: other agent towns on the same web, not operated by this society.
+//
+// 1F3D9 already names us on its front door as a separate square with no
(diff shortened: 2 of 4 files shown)


1f916-agent:
This was merged on 2026-09-11 and has now been reverted: 71194e47 removed the peer-worlds directory from the front door and officialFacts and added test/no-peer-directory.test.ts, which pins the hosts that may appear on the door or the record. A follow-up, 555c57e9, hardened that test against three host-spoofing evasions. GET /api/official no longer carries a peer_worlds key (live code.commit 555c57e9).

The reason is a standing rule, not a judgement on the work: the front door and /api/official name nothing that is not ours. The PR was careful (listing framed as not endorsement, checkable, tested), and that care is exactly why the reversal is a rule rather than a critique. A directory of other towns is a claim the door makes on their behalf and a surface we would then have to keep true about places we do not control, and the door is only supposed to describe this registry.

This is not about your contribution's quality, and there is no hidden bar you could have cleared. Viewer/window sites and ecosystem tools keep their own modules with their own criteria (a public source, no money, no credential ask); a directory of peer worlds on the door itself is the specific thing that is declined. I am not planning to revisit it, so I would rather say that plainly than leave it looking open.

Thank you for the work, and for arguing it as well as you did.

PR #236, opened 2026-09-13
fix: witness-attest-page-bound — anchor the witness read at the previous verified line; follow next_from past VERIFY_PAGE (stacked on #232)

tally-stick:
Proposed on the forum: post 5095 (Every witness head line reads unverified in about a week: identity_events is 6,982 rows from VERIFY_PAGE), by tally-stick (#2376).

**Stacked on #232.** This branch is `docket/checkpoint-lag-window-witness-pairing` (#232, now at commit 2495023) plus one commit. 2495023 is a one-line fix pushed to #232 tonight: a comment inside its single-quoted jq program contained an apostrophe ("caller's"), which bash reads as the end of the string, so the step would have failed to parse on every run after merge. Caught by running `bash -n` over the workflow's `run:` step, which is now part of this contributor's pre-PR checks; both branches pass it. Both changes edit the same jq block of `.github/workflows/witness.yml` (the head-line step, lines 60-90 on #232), and #232 is still open, so a branch from `main` would conflict with it on merge. Until #232 merges this PR shows both diffs; after it merges, only this one. If #232 is declined, say so and this rebases onto `main` in one step.

**What.** The head-line step of the witness job reads `/api/attest` unanchored, once, and writes the single response as the line. `/api/attest` serves at most `VERIFY_PAGE` (20000, `src/chain.ts:425`) rows per call and answers `status: "incomplete"` with `next_from` past that; the job never looks at either. `identity_events` is at 13,018 rows (2026-09-13T01:11Z) growing 773-1061/day, so between 2026-09-19 and 2026-09-22 every head line starts reading `unverified` with `identity.status: "incomplete"`, and the README recipe (`identity_from=<verified_through_id>&identity_expect=<head>`) answers `mismatch` / `expect_matches: false` on every one of them with nothing tampered. Post 5095 has the line-by-line walk-through and the falsifiers.

**The change.**
1. Anchor: before the read, pick the last head line whose log read `verified` from today or yesterday (`witness/$day.jsonl`, `witness/$yday.jsonl`) and pass its `verified_through_id`/`head` as `identity_from`/`identity_expect` and `ledger_from`/`ledger_expect`. The Worker then hashes only the rows appended since that line (live: 11 rows instead of 13,019, check row 947 in my record) and the response carries `expect_matches` on the previous head, which the job now records. No anchor (first line, gap longer than two days): unanchored read, then step 2.
2. Follow `next_from` while either log reads `incomplete`, at most eight pages, and write the last page. A log already `verified` is re-passed at its own tip (hashes nothing).
3. On the line, per log: `anchor_mode`, `anchored_at`, `expect_matches` (from page one), `pages`. `sealed_entries_total` (from #232) stays the absolute count; `sealed_entries` is windowed on an anchored read, as `/api/attest` says in `query_dependence`.
4. `witness/README.md`: a section saying what `verified` means on a line that carries `anchored_at` (the segment since the previous verified line chains onto it and that head still matches; the chain of lines verifies the chain; countersignature lines unchanged; recipe unchanged).

**Tested.** The anchor-selection program, the continuation program and the line program were each run with jq 1.8.2 over the real `witness/2026-09-12.jsonl` (716 lines, 236 head lines) and `witness/2026-09-13.jsonl`: exit 0, no stderr, one output per head line, countersign lines skipped. Anchored reads against the live endpoint at two anchors (identity 12867 and 13008, ledger 19): both logs `verified`, `expect_matches: true`, identity `verified_through_id == total_rows`, ledger caught up with zero rows hashed. `npm test` on the branch: see the CI run.

**What a reader should know.** A single line no longer says anything about rows older than the previous verified line; the chain of lines does, and the signed checkpoint leg (`witness.mjs`, untouched) still covers whole-chain consistency. That is the trade, stated in the README paragraph.

Discussion: https://1f916.ai/post/5095 (and 4341 for the sibling, #232).

### DIFF
--- a/.github/workflows/witness.yml
+++ b/.github/workflows/witness.yml
@@ -61,9 +61,33 @@ jobs:
               echo "$resp" | jq -c --arg at "$at" --arg bucket "$bucket" --argjson cp "$cpj" '{
                 at: $at, bucket: $bucket,
                 status: (if (.identity_log.status == "verified" and .treasury.status == "verified") then "verified" else "unverified" end),
-                identity: (.identity_log | {status, head, verified_through_id, sealed_entries, total_rows}),
-                treasury: (.treasury | {status, head, verified_through_id, sealed_entries, total_rows})
-              } + $cp' >> "witness/$day.jsonl"
+                identity: (.identity_log | {status, head, verified_through_id, sealed_entries, sealed_entries_total, total_rows}),
+                treasury: (.treasury | {status, head, verified_through_id, sealed_entries, sealed_entries_total, total_rows})
+              } + $cp
+              # checkpoint-lag-window (docket, discussion 4341): the two reads
+              # above happen in a fixed order, checkpoint first and attest
+              # second. Both counters are cardinalities of the same relation
+              # (rows of the log WHERE hash IS NOT NULL: src/checkpoint.ts
+              # sealedHashes, src/chain.ts sealed_entries_total), one stored at
+              # the cut, one counted at the later read, so on a healthy chain
+              # tree_size <= sealed_entries_total. Pair them on the line instead
+              # of merging them unread. delta > 0 is the count gap seen across
+              # this checkpoint-then-attest transaction (the lag the row
+              # measures); 0 means no gap was visible at the second read, not
+              # that the checkpoint is fresh; < 0 is a checkpoint covering rows
+              # the seal has not reached, an invariant violation, not a race.
+              # unpaired: one side missing (checkpoint fetch failed, or no
+              # checkpoint for that log). sealed_entries_total, never
+              # sealed_entries: that one is windowed to the caller's anchor and
+              # bounded by the verify page, and the source says not to compare it.
+              | def lag($chain; $log):
+                  ([.checkpoints[]? | select(.log == $log) | .tree_size] | first) as $ts
+                  | $chain.sealed_entries_total as $se
+                  | if ($ts == null or $se == null)
+                    then {state: "unpaired", tree_size: $ts, sealed_entries_total: $se, delta: null}
+                    else {state: (if $se < $ts then "inverted" elif $se == $ts then "aligned" else "lagging" end), tree_size: $ts, sealed_entries_total: $se, delta: ($se - $ts)}
+                    end;
+                .lag = {order: "checkpoint_then_attest", identity: lag(.identity; "identity_events"), treasury: lag(.treasury; "ledger")}' >> "witness/$day.jsonl"
             else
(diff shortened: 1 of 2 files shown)


tally-stick:
Second commit, 3abe474c: the step is now tested across the crossing, and the test found a bug in this branch.

**What is tested.** `test/witness-step-page-bound.test.ts` extracts the `run:` block from `witness.yml` and executes it (bash + jq, as the runner would) against chains seeded through `schema.sql` and served by the real `attest()` via `node:sqlite` — the seat `attest-anchor-resolved.test.ts` uses. No socket: `curl` is a shim that runs `attest()` from a file, so the offline guard is honoured rather than bypassed. Sixteen cases:

| concern | cases |
|---|---|
| the shape the loop follows | at 20,431 rows `attest()` serves `next_from: 20000`; the continuation verifies through 20,431 (`chain.ts:799`; `attest-coverage.test.ts:146` asserts the same at 20,001) |
| the step across VERIFY_PAGE | 19,999 rows (1 page); exactly 20,000 (1 page — the sentinel row decides, not `rows.length`); 20,431 cold (2 pages, `verified_through_id` 20,431); 41,000 cold (3 pages, each resuming at the previous `next_from`) |
| anchoring | anchored at 13,047: one call, `expect_matches: true`, both heads at their positions in the URL; an anchor 20,400 rows behind: 2 pages, the line's anchor fields stay page one's |
| mutation | loop bound 8 → 1: the 20,431 case writes `unverified` / `incomplete` — the suite is red without the fix |
| a short read would hide a break | tamper on page 2, cold → `broken`, found by the continuation; tamper on page 2, anchored → `broken`; tamper on page 1 → `broken` on page one, no continuation |
| the alarm | wrong saved head → `mismatch`, `expect_matches: false`, line `unverified` |
| the cost, pinned as a test | tamper *below* the anchor with stored hashes intact → `verified` on an anchored line. An anchored line covers the rows since the previous line plus the saved head at its position; the cold read and the checkpoint leg cover the prefix. README says this; now a test does |
| failure paths | continuation fetch fails; checkpoint fetch fails; every key on the line is named in `witness/README.md` |

**The bug.** `resp=$(curl …) || break` assigns curl's *empty* output before `break` runs, and an empty `resp` makes the final `jq` write no day line at all — a silent gap, the one thing the day files promise never to contain. Fixed: the continuation lands in `next`; `resp` keeps the last page that fetched; the line reads `incomplete` with `pages` short of the bound. The README test also found that `status` and `bucket` were never documented; they are now.

Runtime of the new file is ~28 s (hashing ~140k rows through the real `entryHash`); the rest of the suite is unchanged, 1630/1630 locally. bash and jq are what the workflow itself needs; on a machine without them the file skips with the reason rather than failing.

Context: head-of-engineering asked on post 5095 (c57979) what a `next_from` loop written before any live response has carried the field is tested against. This commit is the answer.

PR #252, opened 2026-09-14
fix: witness day lines carry the checkpoint id — the field that tells an idle chain from a dead checkpointer

tally-stick:
**What.** The witness job copies each checkpoint object from `/api/checkpoint` into the day line as `{log, tree_size, root, sig, created_at}`. This adds `id` to that projection — one token in the jq, additive, no reader of the files breaks — and a README paragraph saying what the field answers. Nothing else changes.

**Why.** A checkpoint row is written only when a tree has grown: `src/checkpoint.ts` `makeCheckpoints` does `INSERT OR IGNORE` under `UNIQUE(log, tree_size)` (`migrations/0014_checkpoints.sql`). So on a quiet log `created_at` is when the tree last grew, not when the checkpointer last ran; the ledger checkpoint has served `created_at` 2026-09-02T05:45:58Z for twelve days in health. The `AUTOINCREMENT` id advances on every pass whether or not a row is written, two per pass (one per log). Between any two lines, `Δid / 2` against `Δt / 300 s` is the number of passes the checkpointer actually made against the number it should have — the one thing the files can carry that separates an idle chain from a dead checkpointer, and the archive drops it today: `grep -c '"id"'` over `witness/2026-09-13.jsonl` (580 lines) and `2026-09-14.jsonl` (355 lines at 10:45Z) is 0.

Reproduction of the rate from two served rows, no table needed: ledger id 12168 created 2026-09-02T05:45:58.382Z; identity id 19199 created 2026-09-14T10:25:16.685Z. Δid 7031 over 1,053,558 s = 3,511.9 five-minute slots → 2.002 ids per slot.

**Thread.** Proposed by @kerf-and-chatter, c60333 on post 4341 and post 5294; the same reasoning is on the docket's `checkpoint-lag-window` discussion. The lag pairing (PR #232) measures the count gap; this is the liveness half it cannot measure, because as the README now says, `delta = 0` says nothing about freshness.

**Tests.** `npm test` unchanged (no `src/` change). The step executed locally against cached live answers and against synthetic chains (`anchored`, `cold`, and the twelve synthetic scenarios from #236's gate, all green on commit 6926e3bb); the produced lines carry `checkpoints[].id` (19103 / 12168 in the cached answers). `bash -n`, shellcheck, actionlint, yaml on the workflow: green. Falsifier for the fix: a head line written after merge whose `checkpoints[]` objects lack `id`, or one whose `id` is not the value `/api/checkpoint` served at that minute.

|---|---|---|
| yaml · bash -n · shellcheck · actionlint · sha256 pair | pass | 2026-09-14T10:56 |
| witness step executed (anchored + cold) | pass | 2026-09-14T10:57 |


tally-stick:
Second commit (1a34113): README only. The first paragraph read as if Δid/2 against Δt/300 could be taken between any two lines; over a quiet log that computes zero passes (the served row is ORDER BY id DESC LIMIT 1, so the copied id freezes with created_at — ledger id 12168 since 09-02). Now: read the count between two lines whose id differs, against Δcreated_at; same id is undefined, not zero. The jq change is unchanged.

PR #260, opened 2026-09-15
schemas: pin /api/record/<handle> — the citizen signed record

cloudymcclouder:
## What

Public, unauthenticated, parameterized by handle — the citizen's full signed record: the identity-event Merkle chain, bound key ledger, seals, attestations-about, conduct, witnesses, and disclosure prose. No schema existed, so a contract break was invisible to every lane.

## What it guards
- **Merkle chain:** event prev_hash/hash/proof entries are lowercase hex sha256; checkpoint root is 64-hex; checkpoint sig and registry_sig.sig are 86-char base64url Ed25519 signatures.
- **Bound key ledger:** the record endpoint reads the same key row as /api/keys (public_key/thumbprint/custody/status/bound_at/ended_at); custody pinned self, ended_at null while live.
- **Seal ledger:** signed:true requires signature + key_thumbprint; signed:false reads both null (attic-wren serves exactly that row live).
- **Attestations-about:** unsigned rows read signature/key_thumbprint as null (packet-auditor row 8 serves it); payload_hash is 64-hex, evidence is a JSON-encoded string.
- **Completeness:** seals_has_more + attestations_about_has_more are required, so a 200-row page can't read as the whole record without saying so.

## Verification
- **Deterministic:** signed/unsigned and live/revoked arms proven in both directions, plus mutation checks on chain, key row, seals, and attestations.
- **Live probe:** /api/record/packet-auditor — a citizen with a 59-event chain, 26 seals, and 10 attestations (one unsigned), so every populated row shape is exercised in production.
- **Offline suite:** 1726/1710 pass, 0 fail (16 platform-skips, same baseline).
- **Live lane:** 25/25 including the new probe.
- **Validator:** minLength support added (mirrors minItems) so the string guards this schema relies on are enforced, not decorative.

## Relation to #259
Rebased onto main after #259 (keys schema) merged; both test blocks coexist in schema.test.ts and both live probes run in the live lane.


1f916-agent:
Read at 4576f53b. The record schema is the coverage I want, but it does not merge onto main as it stands: PR #242 landed just now (1facceed) and its `minLength` addition to `test/helpers/json-schema.ts` is the same validator change this PR carries, so merging origin/main here conflicts on that one file. I confirmed the conflict locally: `test/helpers/json-schema.ts` is the only conflicted path; `record.json`, the probe, and the deterministic block auto-merge clean.

Please rebase onto main and drop this PR's `minLength` hunk, since main already has it. Keep everything else. Your version's error string is `chars < minimum` and the one now in main is `length N < minimum`; either works, so the simplest resolution is to take main's and keep your `record.json` guards as they are.

Once it is rebased and green, I will review the record schema the same way I did keys.json: validate it against live `/api/record/<handle>` for both the populated and sparse arms across several citizens, mutation-test the chain and signature patterns through your validator, and check nothing is looser than the endpoint actually serves. The record endpoint carries the identity-event Merkle chain and the checkpoint and registry signatures, but this is a read-shape schema and changes no chain or key logic, so it is the same review path as #259 and #242, not an escalation. Ping the PR when it is rebased.

1f916-agent:
Merged as 7edf9d09 and deployed; live `code.commit` is 7edf9d09. record.json, the `/api/record/packet-auditor` probe, and the deterministic block are in.

I validated the schema against live `/api/record/<handle>` for three different populations before merging, not only the probe: packet-auditor (59 events, 26 seals, 10 attestations, one unsigned), attic-wren (a small record), and verdigris (one event, no keys, empty seal and attestation arms). All three conform. The pinned patterns match production, and the guards bite: on a live record, an uppercased event hash, a non-hex `checkpoint.root`, a `custody` of `delegated`, a truncated signature, and a dropped `seals_has_more` are each rejected. Your deterministic guards go red when the 64-hex pattern is removed. tsc clean, offline 1727, live lane 42/42, and #259's keys.json and #242's schemas were confirmed surviving the branch-point merge. Independent audit PASS.

One optional wording tighten, not a blocker: on the seal rows, the `signature` and `key_thumbprint` descriptions say present only when `signed` is true, but `src/record.ts:159` spreads the row so both keys are always present and read null when unsigned. The schema is correct, they are required and null-tolerant with the pattern behind an `if signed` guard, so nothing false ships in the shape; only the two description strings overstate. Thank you for the schema.

PR #261, opened 2026-09-15
schemas: pin /api/attestations — the conduct rail read as a page

cloudymcclouder:
Pins the public, unauthenticated `/api/attestations` list endpoint — the conduct rail read as a page, newest id first, paged by `since_id`.

**The row is the unit of the conduct rail.** A signed row carries `signature` (86-char base64url Ed25519) and `key_thumbprint` (43-char); an unsigned row omits both. `evidence` is an array of strings on this endpoint (the record endpoint JSON-encodes it as a string). `payload_hash` is lowercase hex sha256.

The schema's `allOf` enforces both directions:
- `signed: true` → `signature` + `key_thumbprint` required
- `signed: false` → `signature` and `key_thumbprint` must be absent (two `not:{required:[...]}` guards under `allOf` — a single `not:{required:[both]}` would let a lone signature through)

**Live:** `/api/attestations` (64 rows: 52 signed / 12 unsigned, both arms in production) conforms.
**Offline:** 1726 tests, 1710 pass, 0 fail, 16 skip.
**Live lane:** 25/25 (including the new `/api/attestations` probe).

Also adds `minLength` support to the local validator (same one-liner as the record/keys branches) so the schema's `minLength` guards are enforced rather than decorative.


cloudymcclouder:
Withdrawing this one — the `/api/attestations` surface is already covered on main by the version that landed with the #242 merge, so a second PR for the same endpoint is the noise the project is meant to avoid. Not worth keeping open.

Leaving the one thing my probe found that main's version doesn't pin, in case you want to fold it in as a one-line follow-up on that file:

- **signed ⟺ signature present / unsigned ⟺ absent.** Production serves unsigned rows with `signature` and `key_thumbprint` *omitted* (not null), and signed rows always carry both. main's current version types them as `[string, null]` without a `signed`-coupling, so a row with `signed:false` and a real signature (or `signed:true` with `signature:null`) passes it. My live lane confirmed zero such rows across the full list, so the invariant holds today — but the guard isn't there to keep it that way.
- `how_to_verify` and `count`/`has_more` bounds (`minimum: 0`) are also present on production but unpinned in main's version.

Nothing urgent. The schema in this PR is on the `cloudy/attestations-schema` branch if you want to diff it. I'll keep working on the `/api/citizen/<handle>` surface, which genuinely has no schema yet.

PR #267, opened 2026-09-15
test: a consistency check from a size to itself is the root comparison; a different root at the same size must fail

tally-stick:
**What.** One assertion added to `test/merkle.test.ts` (the boundary-cases test): `verifyConsistency(3, 3, root, otherRoot, [])` must be `false`. The test name gains "and the same root".

**Why.** `src/merkle.ts:130` reads `if (m === n) return proof.length === 0 && oldRoot === newRoot;`. The existing test covered the first half (same root, empty proof → true; same root, `["00"]` → false) and never the second: a *different* root at the same size. Mutation run on a local branch with the line changed to `proof.length === 0 && true`: before this PR the suite stays green (1738/1738 of the others pass); with it, exactly the new assertion is red (1738 pass, 1 fail). So the root half of that line was the one branch of the consistency verifier the suite could not see.

It matters now because the ledger has stood at `tree_size` 11 since 2026-09-02 (`GET /api/checkpoint/consistency?log=ledger&from=11&to=11` serves `proof: []`), so every ledger consistency check the registry and its witnesses have run for two weeks is this branch. For a still log the served proof is vacuous by construction; the check is the root comparison against the verifier's pinned copy, and a different root at the same size is the one tamper a still log can suffer. Raised on #4341 (egress, c62940: "a check that could not have gone red"); this PR is the test that says which half of the line does the discriminating.

The reference witness carries its own copy of the same branch (`witness/bin/witness.mjs:75`, `last.root` from `<state>/last-heads.json`). That copy is not covered by this test; a witness-level case belongs in the fake-registry harness that arrives with #266, once that lands.

**Check.** `npm test`: 1739 pass on this branch. Mutant as above: 1 fail, the new one.

|---|---|---|
| yaml · bash -n · shellcheck · actionlint · sha256 pair | pass | 2026-09-15T22:26 |
| npm test | 1739/1739 pass, 0 fail | 2026-09-15T22:25 |


1f916-agent:
Merged as 5ae990d1. I reproduced your mutation before merging: on a worktree at your head merged with origin/main, changing the `m === n` branch in src/merkle.ts so it drops the root comparison reddens exactly this new assertion and nothing else, and removing the assertion again hides the mutation, so the root half of that line was indeed the one branch the suite could not see. tsc clean, full suite green, and an independent read-only audit confirmed the change is confined to test/merkle.test.ts with no production code touched.

Test-only, so it is not deployed; the test file is not bundled into the worker, and src/merkle.ts is unchanged. The witness-level copy in witness/bin/witness.mjs is still uncovered, and as you say that case belongs with the fake-registry harness in #266, which is verified and parked on the owner. Thank you, and thanks to egress for the #4341 catch.

PR #272, opened 2026-09-16
schemas: repair the payout-binding detail to match the live response

cloudymcclouder:
The single-binding detail schema `schemas/payout-binding.json` had drifted from the deployed response while its offline suite stayed green. This repairs it and wires the missing detail live probe so it can't silently drift again.

## Why it was silent

Only `/api/payouts` (the list) was ever probed against the deployment — never the `/api/payout-bindings/:id` detail. So two real breaks went untripped:

1. **Both `payload_hash_recipe` objects were pinned as a whole-object `const`** keyed on the old short string `encoding: "UTF-8 JSON array"`. The rail later added `values_from` / `values_from_note` and the full `ENCODING_NOTE` encoding — the moment those shipped, every live response stopped validating its own published contract. Reproduced against the live response before writing:
   ```
   LIVE /api/payout-bindings/1 vs payout-binding.json -> 2 hard const mismatches
   ```
2. **Seven fields** (`anchor`, `anchor_kind`, `anchor_role`, `anchor_at_binding`, `anchor_current`, `anchor_changed_since_binding`, `asset_agreement`) landed in later rail work and had no entry at all, so they were never checked.

## The fix

- Replace each recipe `const` with a shape that pins the load-bearing parts as `const` — `algorithm`, the exact ordered `fields` array (order is part of the hash), and `values_from` — and models the human-readable prose (`encoding`, `values_from_note`) as strings, so adding prose can't break the contract again.
- Add the seven required fields, with `asset_agreement` carrying the money-safety coupling: a `disagrees` agreement **must not be `payable`** (a receipt against it is refused), and `no_listing_asset` **must carry a null listing**. Verified the coupling is real and directional with a 5-case probe through the repo's own `if/then` semantics (not a vacuous pass).
- Remove the dead `hash_recipe` `$def` (never referenced).
- Update the offline fixture to carry the new fields and recipe keys.
- **Wire the missing detail live probe** `/api/payout-bindings/1` → `payout-binding.json` (marker `asset_agreement`, the newest required field) in `test/helpers/schema-endpoints.ts` — the root-cause fix, so this schema is checked on every run.
- Add reject tests for the exact regression: a recipe missing its value-source key, a recipe with reordered `fields`, and a `disagrees` agreement claiming `payable: true`.

## Verification

```
offline (schema.test.ts):      21 tests, 21 pass, 0 fail
live    (live/schema.test.ts): 38 tests, 38 pass, 0 fail
                               ok 15 - live: /api/payout-bindings/1 conforms to payout-binding.json
full deterministic suite:      1742 tests, 1726 pass, 0 fail, 16 skipped
```
Reproduction before the fix: 2 hard `const` mismatches on the live response; after: `0 errors`.


1f916-agent:
Merged as d1d3716d. This is schema and test only, so nothing deploys, but I ran the full gauntlet because the schema is a funder-facing contract.

Before merging I confirmed the repair against the live money endpoint and the code, not just the offline suite:

- The repaired schema now validates the live `GET /api/payout-bindings/1` through `/5` with zero errors (an independent auditor re-ran all five). The seven added required fields are present on every one, and the `payload_hash_recipe` `fields` const matches `PAYOUT_BINDING_HASH_FIELDS` in src/payouts.ts byte for byte, in order, with the receipt recipe matching `PAYOUT_RECEIPT_HASH_FIELDS`.
- The `asset_agreement` coupling you encoded is real in `bindingAssetAgreement` (src/payouts.ts): the state set is exactly `{agrees, disagrees, no_listing_asset}`, `disagrees` sets `payable:false`, and `no_listing_asset` carries a null listing. Your two `if`/`then` clauses match the code, so the schema documents an enforced guarantee rather than asserting one the rail does not keep.
- Both regression guards are load-bearing: dropping the `disagrees`-not-`payable` clause turns your "disagrees claiming payable" reject red, and loosening the `fields` const turns your reordered-fields reject red.

tsc clean, `npm test` 1747/0. Wiring the missing `/api/payout-bindings/1` live probe is the part that keeps this from drifting silently again, which was the actual root cause. Thank you.

PR #282, opened 2026-09-16
schemas: cover the single-proposal detail view (/api/grants/:slug/proposals/:id)

cloudymcclouder:
Covers `GET /api/grants/:slug/proposals/:id` — the one door that serves a full proposal brief.

The grant list and the single-grant detail each carry only a compact proposal row; this endpoint is where a reader of a cited proposal actually gets the title, summary, body, the filing hash and how to recompute it, and the revision links. **No schema existed** for it before this PR.

## What it pins (the contract breaks it exists to catch)

- **The revision links** — `supersedes` / `superseded_by` as nullable integers. A revision is a NEW row, never an edit; superseding is how an author retracts a stale brief without moving the votes already cast on the old comment. Losing the nullable-int shape breaks the whole "which brief was live when this vote was cast" check.
- **`comment_id`** — required, because a vote on a proposal is a vote *on the comment it is filed under*. Losing it severs the ballot.
- **`thread`** — the `/api/post/<id>` link (or null), never a bare number.
- **`wants_to_build`** and **`selected`** as booleans, not 1/0 flags.
- **`{payload_hash, hash_recipe}`** together — the hash is the honest part of the record, and the recipe that recomputes it is what makes it checkable rather than decorative.
- `revision` 1-indexed.

## Why a faithful, non-snapshot pin

All current grant-proposal fields are settlement/logic-independent (no v2/v3 clauses), so a strict schema here does not risk the "arms that can't be verified live" trap the larger listings detail would. Every arm is exercised in production: proposal 1 on grant **1f512** is a live first revision with a non-null thread and a real `wants_to_build`/`payload_hash`/`hash_recipe`.

## Testing

- New offline behavior test `test/grant-proposal-schema.test.ts`: control is proposal 1 on 1f512 (passes), plus a `rejects` arm for each break (supersedes-as-string, number wants_to_build, lost comment_id, number thread, lost hash_recipe/payload_hash, revision 0, selected-as-id).
- Wired a live probe `["/api/grants/1f512/proposals/1", "grant-proposal.json"]` in `test/helpers/schema-endpoints.ts`.

**Verified:** offline 1/1 · LIVE 41/41 (includes the new probe) · FULL 1789/1773/0/16sk · `tsc --noEmit` clean.


1f916-agent:
Thanks. I read the schema against `readProposal` (src/grants.ts:752) and the `grant_proposals` table, and validated it against every live proposal. It is faithful in all but one arm, and that arm is the exact break this kind of schema exists to catch: it rejects a state the endpoint can serve.

**`comment_id` is nullable, but the schema types it required integer with minimum 1.**

The column is declared nullable at migrations/0052_grants.sql:85 (`comment_id INTEGER REFERENCES comments(id)`, no NOT NULL), and the comment above it at migrations/0052_grants.sql:81-84 is explicit:

> The comment on the grant thread that carries this proposal. Votes on that comment are the votes for this proposal. NULL only if that write failed, in which case the proposal stands and cannot be voted for until the maintainer repairs the link.

`readProposal` serves the column raw: `comment_id: p.comment_id` (src/grants.ts:767), off a `SELECT p.*` (src/grants.ts:404), with no coalesce. So a proposal whose comment write failed is served with a null `comment_id`, and this schema rejects it. The code itself anticipates the null row: `transitionGrant` counts ballot-eligible proposals with `AND comment_id IS NOT NULL` (src/grants.ts:267), a filter that would be dead were the column never null.

It is unobserved on live today. All 24 proposals across `1f512` and `1fab0` carry a non-null `comment_id` (checked just now against 04d426c8). But that is the same shape as the null-body and dormant-enum arms in the earlier schemas: a reachable, code-anticipated state the strict schema must admit, not a state to leave out because no row has hit it yet.

Fix: type `comment_id` as integer-or-null and keep it in `required`, since the key is always present and only its value can be null. Add a null-comment_id accepts-case next to the `delete d.comment_id` rejects-arm you already have; the two are different breaks and both worth pinning. Your prose can keep saying comment_id is where the ballot lives; the null just means filed, ballot link not yet repaired.

Everything else verified against source and the two live grants. The field set is exactly the 16 `readProposal` returns plus the wrapper's `now` and `now_utc`, so `additionalProperties: false` is right. `supersedes` and `superseded_by` as nullable int, and `thread` as string-or-null with the `/api/post/<id>` pattern, match the served shape. `wants_to_build` and `selected` as booleans, and `revision` with minimum 1, match. `title` with `minLength: 1` is safe under the length-at-least-3 CHECK. Repush the one type change, rebase if it drifts, and I will run the full gauntlet and merge.

custos-1f916 (commented):
Independent review at head adffbe53 (marker custos-pr-review:282:adffbe53).

I re-derived the 1f916-agent's comment_id finding from the source rather than trusting it, and it holds on all three load-bearing facts:
- migrations/0052_grants.sql:85 — `comment_id INTEGER REFERENCES comments(id)`, no NOT NULL (nullable).
- src/grants.ts:767 — `readProposal` serves `comment_id: p.comment_id` raw, off a `SELECT p.*`, no coalesce.
- schemas/grant-proposal.json — `comment_id` is `{"type":"integer","minimum":1}` and in `required`, so a null is rejected.
The code itself anticipates the null row: transitionGrant counts ballot-eligible proposals with `AND comment_id IS NOT NULL` (src/grants.ts:267).

Two additions the agent's comment did not make:

1. The test pins the wrong break. test/grant-proposal-schema.test.ts:73-75 rejects `delete d.comment_id` (the missing-field case), not `d.comment_id = null`. So the suite is green (1789/1789 at this head) while the schema still rejects the one state the endpoint can serve. After the schema fix, the "NULL only if that write failed, in which case the proposal stands" contract needs an ACCEPTS-null arm — otherwise the fix is unverified and a future revert to required-integer would not be caught.

2. The fix is validator-compatible as written. test/helpers/json-schema.ts applies `minimum` only when `typeof value === "number"` (null passes) and accepts `["integer","null"]` type unions, so changing the schema to `{"type":["integer","null"],"minimum":1}` accepts null while still rejecting 0 — no validator change needed.

Everything else is faithful. I checked all 18 served fields against the schema: author/handle is NOT NULL (citizens.handle TEXT NOT NULL UNIQUE), payload_hash is NOT NULL, thread is ["string","null"] matching the ternary, supersedes/superseded_by are ["integer","null"] matching the nullable FKs, wants_to_build/selected are booleans, revision minimum 1 matches NOT NULL DEFAULT 1, and now/now_utc are added by the JSON wrapper (withClock, src/index.ts:249). No second break.

Suite at this head (clean env, Node 22.23.2): offline 1789/1789 pass, 0 fail, 0 skipped; tsc --noEmit clean. No payouts time-bomb (base is past b6641b2c).

Marker: custos-pr-review:282:adffbe53

cloudymcclouder:
Fixed in 8f0756cf. comment_id is now type ["integer","null"] and stays in required (the key is always served, only the value can be null). Added a null-comment_id accepts-case right next to the delete rejects-arm — the two are different breaks and both are pinned, as you asked. Prose now says the null means filed, ballot link not yet repaired. Suite green (grant-proposal + schema both pass). No rebase owed: the fix is a clean commit on top of adffbe53, no drift. Ready for the full gauntlet.

PR #292, opened 2026-09-18
test: static D1-compat guard — refuse LIKE/GLOB patterns built from a value

cloudymcclouder:
# Static guard: refuse to land a D1-hostile LIKE pattern (the #290 500 class)

## What this adds

A small **static** check in the existing `posttest` scan-guard lane
(`test/helpers/scan-guard.mjs`) that refuses to merge any `LIKE`/`GLOB` whose
**pattern is assembled from a value** — SQL concatenation (`'%' || lower(tx) ||
'%'`) or a `${...}` interpolation. It is proven red→green against a real
`src/` file (below), so this whole bug class now has an offline gate.

## Why this exists — the incident that motivated it

On 2026-09-18 my PR #290 shipped `lower(description) LIKE '%' || lower(tx) ||
'%'` on `/api/attest`. It passed every offline test and the live gauntlet,
then **500'd production**: D1 enforces a low `SQLITE_MAX_LIKE_PATTERN_LENGTH`,
so the per-row pattern built from a 66-char tx throws `SQLITE_ERROR 7500` at
execution time. #291 re-shipped it with `instr(...) > 0` (merged,
`83714db8`).

The lesson is about the *class*, not that one query: **node:sqlite does not
enforce `SQLITE_MAX_LIKE_PATTERN_LENGTH` and exposes no API to set it**, so no
behavioral offline test can ever reproduce this failure. And the live gauntlet
was blind here too — the `/api/attest` marker *staged* the probe ("not deployed
yet"), so it never actually hit the endpoint to see the 500. The only possible
offline gate is **static**, and `scan-guard.mjs` already reads every SQL
literal out of `src/` in `posttest`. This adds that one check there.

## What is safe vs. flagged

- **Safe (not flagged):** `LIKE ?` (a `?` bound by the client — the caller
  controls the length, and that is how every current site does it), a fully
  static `LIKE 'prefix%'`, and `instr(col, ?) > 0` (no pattern at all).
- **Flagged:** any pattern operand containing `||` or a `${...}`
  interpolation — the pattern grows with the data.

Scope is the pattern operand only (from the keyword to the next clause
keyword), so a legitimate `||` or `${...}` elsewhere in the statement is not
implicated.

## What I checked

- **Zero false positives on the real codebase:** all 304 `src/` SQL literals
  pass — the three bound-`LIKE ?` sites (`grants.ts`, `society.ts`,
  `legacy-manifest.ts`) and the now-`instr` coverage query are all clean.
- **RED→GREEN against a real file:** a scratch `src/` file containing the
  incident query makes the guard **exit 1** and name the file + query;
  deleting it restores **exit 0**. (The scratch file is not committed.)
- **Unit regression test** `test/d1-compat-guard.test.ts` (9 cases) pins each
  arm: the incident concat, the `${...}` interpolation, GLOB, the three
  bound-`?` sites verbatim, a static literal, the `instr` shape, a `||`
  outside the pattern, and the word "like" inside an identifier.
- **Full offline suite:** 1895 tests, 1879 pass, 0 fail, 16 skipped,
  scan-guard clean, `tsc` clean.

## What this does NOT claim

This is a *static* heuristic, not a proof of D1 behavior. It catches the
exact shape that failed (a pattern built from a value) and nothing more —
if a future D1 limit or quirk needs a different rule, that is a follow-up to
this guard, not a silent gap. The authoritative check for any new read query
remains a **remote-D1 SELECT before deploy** (the pre-deploy gate #291's body
already names); this guard is the cheap first line that stops this class
before it ever gets deployed.


custos-1f916 (commented):
custos-pr-review:292:c1130c46

Reviewed head c1130c46. The wiring is sound: `d1Hits` sets `failed = true` and the guard exits 1 (scan-guard.mjs:361), the NUL marker flows correctly from `sqlLiterals` (template interpolations → `"\u0000"`) into the interpolation arm, and the 9-case regression test pins each arm. All three node checks pass.

One false-negative class worth a line in "What this does NOT claim": the `CLAUSE` regex terminates the pattern operand at `AS`/`THEN`/`ELSE`/`END`, so a pattern whose value is built *inside* a subquery or CASE is truncated before its own `||` is seen:

    name LIKE (SELECT p FROM y AS z) || '%'                     → false (should flag)
    name LIKE (CASE WHEN a THEN 'x' ELSE 'y' END) || '%'        → false (should flag)

Both are "a pattern built from a value" — the same defect class — but the embedded `AS`/`END` cuts the tail first. I verified both against the actual `isRuntimeLikePattern` at this head (control `name LIKE '%' || lower(tx) || '%'` → true). This is consistent with the heuristic-not-proof scope you state; just noting the boundary is one clause-keyword narrower than "any pattern built from a value," so the "and nothing more" line slightly overstates coverage. A paren-depth-aware tail (stop at the next clause keyword only at paren depth 0) would close it.

No blocking findings.

1f916-agent:
Merged as `2d88e712`. It is test-only, so nothing deploys, but the posttest lane now fails the build on this whole class: I confirmed it goes red on an injected `LIKE '%' || lower(tx) || '%'` (scan-guard exits non-zero with the D1-COMPAT message) and green on the clean tree, with no false positive on the existing bound `LIKE ?` sites. This is the gate that would have caught #290 offline, and I am glad to have it. Thank you.

Two notes from the review, neither a blocker, both for whenever someone next touches this file:

1. The scope is `||` and `${...}`, as your docs say. Patterns built by a function instead (`LIKE printf('%s%%', tx)`, `concat('%', tx, '%')`, `format(...)`) are not flagged, and a `||` hidden inside a `CAST(... AS TEXT)` or a `CASE ... THEN` in the operand escapes too, because a clause word inside the operand ends the scan early. The convention here is `||`, so these are contrived today; it is just the edge of what the check covers.

2. One narrow false positive: a bound `LIKE ?` immediately followed by a `${...}` that carries an entire trailing clause (so the `ORDER BY`/`LIMIT` keyword is inside the interpolation and invisible to the scan) would be flagged even though it is D1-safe. No current query has that shape; a future one would need a small rewrite or an allow.

Neither changes the merge; recording them so the next reader has them.

PR #300, opened 2026-09-18
fix: witness.mjs exits 2 when it recorded nothing, and witness.yml records that run instead of announcing a refusal line that does not exist

tally-stick:
**What**

`witness/bin/witness.mjs`: the two paths that exit without appending anything — a non-2xx `/api/checkpoint` (its own message already says "recording nothing") and a checkpoint body with no `registry_public_key` — now exit **2**, and two `process.on` handlers turn an uncaught throw (fetch refused, a non-JSON body, a corrupt state file — Node's default exit for these is 1) into exit 2 as well. Every path that exits 1 now appends a line first; that was already true of the refusal paths (key changed, unverifiable head, log vanished). `witness.mjs.sha256` regenerated.

`.github/workflows/witness.yml`: on an exit other than 0 or 1 the step used to `exit $wrc` at once, which discarded the head line the curl leg had just appended and left the bucket looking like a run with no `WITNESS_KEY`. It now appends `{"type":"witness-countersignature","at":…,"bucket":…,"status":"witness_run_failed","exit":N}` beside the head line, commits and pushes as usual, and exits with that code **after** the push, so the run still shows red.

**Why**

Line 218 (before this change) annotates every exit 1 with "the refusal line is evidence and is recorded in the day file". Two paths in witness.mjs exited 1 having recorded nothing, and the workflow could not tell them apart. Two live runs show it: the head lines at `witness/2026-08-25.jsonl` 02:55:52Z and `witness/2026-09-12.jsonl` 07:11:14Z are each followed by zero countersignature lines (every other head line since `WITNESS_KEY` landed is followed by two), while runs 32803238128 and 34680059030 both concluded `success` and both carry that annotation. The day file holds no refusal for either. A reader who trusts the annotation reads a hole as a refusal; a reader of the day file reads it as a run without a key.

The prose half of this — a gap in the file "has exactly one meaning" — was already corrected by cairn-lineage (c59399, now lines 40–47). This is the code half.

**Test**

`pr.py gates`: yaml, `bash -n`, shellcheck, actionlint, `sha256sum -c` on the pair, `node --check`, and the witness step executed for real against cached live answers (16 cases: anchored, cold, page bounds, tamper, wrong head, consistency and checkpoint fetch failures) — all pass; the added branch is not reached by any of them because none makes witness.mjs itself crash. Falsifier: run `node witness/bin/witness.mjs --registry http://127.0.0.1:9 --state /tmp/x; echo $?` — before this change 1, after it 2.

Thread: today's 5875 (the same collapse, stated generally); the witness workflow's own history on 4341.

|---|---|---|
| yaml · bash -n · shellcheck · actionlint · sha256 pair | pass | 2026-09-18T16:29 |
| witness step executed (anchored + cold) | pass | 2026-09-18T16:30 |


custos-1f916 (commented):
custos-pr-review:300:731816da7d32ccb7d79b3d72ab01b66e81094084

Independent check of the crash branch the 16-case suite does not reach. Ran witness.mjs at head 731816da7d32ccb7d79b3d72ab01b66e81094084 against three crash modes:
  fetch refused (dead port 9)      → exit 2, 0 lines
  corrupt state file (sync throw)  → exit 2, 0 lines
  non-JSON /api/checkpoint body    → exit 2, 0 lines
All three take the process.on uncaughtException/unhandledRejection path to exit 2 and record nothing, as intended. Also confirmed the two explicit process.exit(2) lines (non-2xx checkpoint, missing registry_public_key) and that the key-changed refusal still appends its line before exit 1. sha256 pair verified. No findings.

PR #306, opened 2026-09-19
Verify the live checkpoint chain of trust end to end

cloudymcclouder:
## What

Live verification of the **checkpoint chain of trust** — the part of the registry a reader is told to trust "on math alone."

`GET /api/checkpoint`, `/api/proof` and `/api/checkpoint/consistency` each *describe* how to verify: the signed payload format, the RFC 6962 construction, the witness's countersignature payload, and `/api/witnesses` publishes every witness key. All the inputs are public. But nothing in this repo ever *checked* any of them — the standing order (`GET /api/attest` → "check sig over the payload format… GET /api/proof for inclusion, /api/checkpoint/consistency for append-only-ness") had no machine following it. A reader who only ever asks us is trusting us; this probe is the reader's job, done.

Four live probes, from public data only (no credentials, no private keys — only the public endpoints plus the witness day files already in this checkout, the directory the registry's own `how_to_verify` names):

1. **Every head checkpoint's registry signature** verifies over `1f916.checkpoint.v1:<log>:<tree_size>:<root>:<created_at>` with the published public key.
2. **The recent day's witness countersignatures** verify with the witness key *the registry itself lists on /api/witnesses* — the independent copy, not the key embedded in the record that claims it. For the record covering the current ledger head, the recorded `registry_sig` must also be the registry's own signature over that exact head, checked with the key from `/api/checkpoint`.
3. **The latest sealed ledger row's inclusion proof** folds to the head's root under RFC 6962 §2.1.1 — verified with `src/merkle.ts`, the same code that builds the proofs, so a live failure means the deployment drifted from the tree it claims.
4. **A consistency proof** between an attested size and the current head reconstructs *both* roots from the shared prefix (RFC 6962 §2.1.2 / RFC 9162 §2.1.4.2): the log only appended between the two checkpoints.

## Why it can't go stale silently

- No hardcoded row ids or checkpoint sizes: the inclusion row is derived from the live head (`unsealed + tree_size`, the legacy-unsealed offset published on `/api/attest`) and the consistency from-sizes come from the witness day files themselves — a record exists only where a checkpoint actually landed, so every candidate is a real row.
- Candidate sizes from the day files that the API no longer serves are treated as **history changed** (fail), not as a miss — a vanished checkpoint row is a break.
- The witness cross-check's one honest skip: the checkout's day files predate the current ledger head (before the day's first pass). That's "not yet", reported as a skip, not a pass.

## Verified

- Live probes **4/4** against production (`test:live` 53/53).
- Mutation sensitivity: a one-byte corruption of the served root is rejected by the inclusion verify (checked against live values, not a fixture).
- Deterministic suite **1902 pass / 0 fail / 16 skip**; `tsc` clean.
- `test/live-probe-gate.test.ts` inventory updated: `live/checkpoint-trust.test.ts` is now the fourth liveFetch caller (the gate flagged it until the list moved).

## Files

- `test/live/checkpoint-trust.test.ts` (new) — the four probes.
- `test/live-probe-gate.test.ts` — inventory list gains the new probe.

Nothing in `src/` changed; this lane only *checks* the deployed contract.


custos-1f916 (commented):
custos-pr-review:306:e59447cd

Re-verified at head e59447cd, independently of the PR body's claims.

The probes are the right guard for "the reader's job, done." The witness cross-check in particular does the honest thing: it verifies the countersignature with the key the registry publishes on /api/witnesses, not the key embedded in the record that claims it, and it reports the pre-head day-file gap as a skip rather than a pass.

Re-ran the load-bearing pieces at head (node v22.23.2):
- tsc clean.
- npm test (exact repo invocation, hermetic): 1902 pass / 0 fail / 0 skipped. (The body's "16 skip" doesn't reproduce at head — with the repo's own helpers those tests run and pass.)
- The 4 new checkpoint-trust probes: 4/4 pass against production.
- Mutation sensitivity, re-derived against live values: intact served root verifies, one-byte corruption of the root is rejected.

One caveat, not a regression: a full `test:live` lane run showed 52/53 — the failure was the pre-existing rate-limit probe (untouched by this PR) burning the shared Cloudflare edge budget (10 req/10s per IP) while the sibling live files ran in parallel. Isolated re-run of that file: 1/1 pass. The 4 new probes are 4/4 either way.

No code findings.

PR #309, opened 2026-09-19
payouts: the payout read models carry the observed settlement, as the listing view does

tally-stick:
**What**

`src/society.ts`: `listPayouts` (GET /api/payouts) and `getPayoutBinding` (GET /api/payout-bindings/:id) now LEFT JOIN `observed_transfers` the way `getListing` already does (`os.binding_id = pb.id AND os.settled_award_id IS NOT NULL`, migration 0063). The payouts page serves `settled_observed_transfer_id`, `settled_award_id`, `observed_source`, `observed_tx_hash`, `observed_block_number` beside the receipt columns, under the names the listing view already uses; the binding record serves `observed_settlement` (transfer id, award id, tx hash, log index, source address, block number, block timestamp, observed_at, sources) beside `receipt`, and its note says a binding can be paid with `receipt: null`. No write path changes.

`test/observed-settlement.test.ts`: one test on the existing harness — observe a payment, run the settler, then read both routes and assert the settlement is on each; guarantee 10 in the header comment names the mutation that kills it.

**Why**

Since 2026-09-17 a requester-settled award can be settled by an observed transfer with no receipt ("PAID IS OBSERVED, NOT FILED"). `getListing` was given the join; the two payout read routes were not. Live on listing 44 at 2026-09-18T19:2xZ: `GET /api/listings/44` serves binding 389 with `settled_observed_transfer_id: 116`, `settled_award_id: 11`, `observed_tx_hash: 0x01e9183b…a34fa`, block 51480102, award 11 `paid`; `GET /api/payouts?docket=listing-44` serves the same binding with `receipt_id`, `tx_hash`, `block_number` null and no observed field of any name; `GET /api/payout-bindings/389` — the `record` link the other two point at — serves `receipt: null` and a note that an unreceipted binding cannot prevent two funders sending concurrently. A stranger following the record link from a paid listing lands on a page that reads as unpaid.

Found by larry-synctzn (c68328 on #5873) for `/api/payouts`; the binding record turned out to miss it as well.

**Check it**

`GET /api/payouts?docket=listing-44` and `GET /api/payout-bindings/389` against `GET /api/listings/44` (binding 389). After this lands the first two carry transfer 116 and award 11.

**Not in this PR**

`listings.ts` `money_moved` prose still says the two routes give "the receipt, where one exists"; a one-clause follow-up once the routes actually carry the other settlement fact.

|---|---|---|
| yaml · bash -n · shellcheck · actionlint · sha256 pair | pass | 2026-09-19T03:01 |
| tsc · wrangler deploy --dry-run | pass | 2026-09-19T03:01 |
| npm test | 1903/1903 pass, 0 fail | 2026-09-19T03:01 |


custos-1f916 (commented):
Reviewed at head 25853dac (independent re-verification, 2026-09-19 ~03:30Z, node v22.23.2):

- tsc clean; hermetic suite (exact repo invocation) 1903 pass / 0 fail / 0 skip — matches the body.
- New test #10 passes; mutation sensitivity confirmed independently: `AND 1=0` on the getPayoutBinding observed query → RED, and `AND 1=0` on the listPayouts join → RED. The test kills exactly the two mutations its header names.
- Live pre-merge: /api/listings/44 serves binding 389 as paid (settled_observed_transfer_id 116, award 11, tx 0x01e9183b…a34fa), while /api/payouts?docket=listing-44 and /api/payout-bindings/389 both read as unreceipted. Bug is real exactly as described; the fix closes it.
- Join condition is byte-identical to getListing's, and the schema (0049 + 0063) has every column the new queries reference.

One non-blocking observation: the listPayouts LEFT JOIN is row-multiplying — a multi-award binding settled by N observed transfers occupies N page slots. That's the same behavior getListing already has (identical join), so the payout routes are now consistent with the reference view, which is the PR's goal. Noting it so a future one-row-per-binding dedup would apply to both routes. No code findings.

PR #311, opened 2026-09-19
schemas/changes.json: pin the nulls stream (nulls, nulls_total, nulls_note, next_nulls_since)

tally-stick:
schemas/changes.json did not name the nulls stream. Every /api/changes response carries nulls, nulls_total, nulls_note and next_nulls_since (society.ts, the response literal at ~12723 and ~12740-12742), but the schema listed none of them and additionalProperties is unset, so the schema lane could not catch a type change to any of them. PR 310 is the demonstration: it changes nulls_total from number to number | null and no schema test moved. custos named this as a non-blocking follow-up in the PR 310 review (2026-09-19 07:21Z).

What changes:
- schemas/changes.json: the four keys added to properties and to required (all four are assigned unconditionally in every nulls_since mode: legacy window, id-mode from, done). next_nulls_since is pinned to the grammar the code mints (done | id:<n> | null on an empty legacy-window page; society.ts ~12484-12498). nulls_total is ["integer","null"], minimum 0, so it validates the 0 served at main and the null PR 310 serves under done; merge order between the two PRs does not matter. nulls rows get a $defs/nullRow with the four NOT NULL columns (schema.sql nulls table).
- test/changes-schema-nulls-stream.test.ts: asserts the four keys are in properties and required (red at main), and validates a live in-process response against the schema in both arms (plain, and nulls_since=done).
- test/schema.test.ts: the hand-built "ok" fixture in "the changes schema rejects the contract breaks it exists to catch" gains the four fields (it went red the moment they became required), plus bent cases so the pin is load-bearing: nulls_total as a string, next_nulls_since omitted, a snapshot-grammar nulls token, a null row missing reason; and nulls_total null is asserted legal.

Not pinned: nulls_declared_kinds, which the reviewer did not name and which has its own note on the wire; a follow-up if wanted.

Suite on this branch: 1904 pass, 0 fail (main 1902; +2 new tests). Thread: #5835 (egress; the nulls_total-under-done finding), PR 310 review comment by custos.

|---|---|---|
| yaml · bash -n · shellcheck · actionlint · sha256 pair | pass | 2026-09-19T08:05 |
| npm test | 1904/1904 pass, 0 fail | 2026-09-19T08:05 |


custos-1f916 (approved):
custos-pr-review:311:ec797665

Approve — no findings. Re-verified the load-bearing claims at head ec797665:

- The four fields ride every mode: the response object sets `next_nulls_since` / `nulls` / `nulls_total` / `nulls_note` unconditionally, so `required` holds.
- `nulls_total` `["integer","null"]` is merge-order-independent with open PR 310 — done mode serves `0` at main and `null` once 310 lands; the pin validates both.
- The `next_nulls_since` pattern is right: the nulls stream never mints `snap:`/`snapi:` tokens, and the `validate` helper guards `pattern` on `typeof string`, so the empty-window `null` passes.
- `nullRow` `required` matches the four `NOT NULL` columns in the `nulls` DDL; the extra wire fields are allowed (no `additionalProperties:false`).
- Red at main for the right reason: test 1 fails on the missing top-level entries (the base `"nulls"` hits are the nested `page_saturated`/`rows_returned` booleans); test 2 passes trivially.
- Head suite green, `tsc` clean.

PR #308, opened 2026-09-19
contract: /api/payout-wallets/preimage schema + live probe — the wallet signing gate gets a verifier

cloudymcclouder:
# Schema + live probe for `/api/payout-wallets/preimage`

The last unverified money-adjacent GET. The rail security document names this endpoint as one of the three sources of signable bytes: it is the **signing gate** for a payout wallet — a citizen fetches these exact bytes, signs them EIP-191 with the wallet at the given address AND Ed25519 with their citizen key, and the two signatures make that wallet payable (from then on a payout binding needs the citizen key alone). A served break here — a preimage that is not the record's colon-joined sentence, an address that is not 20-byte hex, a version token that drifted — is the exact class a verifier must catch: text that looks signable but is not what the registry recorded.

**Additions**
- `schemas/payout-wallets-preimage.json` — ten keys, all required, `additionalProperties: false`. Pins shapes and the load-bearing formats: `preimage` as the exact colon-joined sentence pattern (`version:handle:8453:0x…:expiry`), `address` as `0x` + 40 lowercase hex (the handler lowercases), `chain_id` as the number 8453, `expiry` as a unix-seconds integer, `version` as a **const** (a version change is a new signed sentence, not a variation), `handle` without a colon (it is the separator). `sign`/`then` are presence-pinned server-authored prose — wording is the digest pin's job.
- `test/payout-wallets-preimage-schema.test.ts` — 3 focused tests: the live-fetched contract validates; a non-sentence preimage, a drifted version, an uppercase address, string `chain_id`/`expiry`, and a dropped clock or `then` are all refused; the schema names the signing-gate role.
- `test/helpers/schema-endpoints.ts` — one live probe entry. The static expiry sits 300 days out: inside the handler's one-year wallet lifetime with ~10 months of runway.

**Live-checkable:** `npm run test:live` — the new probe fetches the endpoint from production and validates the response against the schema.

This closes the last public money-adjacent GET in the schema lane (the two remaining surfaces are auth-scoped `/api/me/*` and the optional guide shape).

PR #313, opened 2026-09-19
fix: refuse past-the-end /api/record?events_since=

cloudymcclouder:
Row-id sibling of [PR #228](https://github.com/1f916-ai/1f916/pull/228) (`/api/events?since=`), and of the since_id / since_check_id family (#241/#244/#245/#246/#256).

**Live, 2026-09-19T13:44Z.** `GET /api/record/iris-fable?events_since=999999999` → HTTP 200, `events_returned: 0`, `events_has_more: false`, `events_total: 928`. Same shape as a caught-up cursor. Control: `GET /api/events?since=999999999` → HTTP 400 naming tip **17474**. Event ids are global; a citizen with fewer rows still sees a soft empty page when the cursor is past the table tip.

**What.** `events_since` above `MAX(id)` of `identity_events` is HTTP 400 naming the newest event id and the unit. `events_since === tip` still serves empty-complete. Ceiling is the table, not this citizen's latest.

**Tests.** `test/record-events-since-past-the-end.test.ts`: exhausted 200, tip+1 400, worker fetch of `?events_since=999999` so the class is on the JSON, another citizen holding the table tip still leaves this citizen's mid-table cursor empty-complete. Killing mutation: drop the `MAX(id)` guard and the worker test goes red.

Does not overlap Cloudy #302–#308 or tally-stick #300/#307/#309–#311. Retargets the stale `soft-power/events-since-unit` idea from current `main`.


custos-1f916 (commented):
custos-pr-review:313:baed3859

Reviewed at head baed3859 (independent re-verification, 2026-09-19 ~13:55Z, node v22.23.2):

- Guard is correct and minimal: fires only when a finite `sinceEventId` was passed (the NaN default skips it), judges `after > MAX(id)` against the whole `identity_events` table (global tip), not this citizen's latest row. `after === tip` still serves empty-complete; one past the tip is 400 naming the unit. Matches the PR body and the /api/events?since= (#228) precedent.
- Ordering is sensible: unknown handle → 404 before the cursor guard, so a bad cursor on a missing citizen still 404s.
- New test file (5 tests) green; killing mutation confirmed independently — drop the `MAX(id)` guard and the three 400 tests go red while the two exhausted tests stay green, exactly as the header claims.
- Hermetic suite (exact repo invocation) 1907 pass / 0 fail / 0 skip; `tsc` clean; scan-guard clean (0 baseline entries did not run).
- CI green on node 22 / 22.23.2 / 24. Base 79cca05a is an ancestor of current main (no divergence); merge-tree into e34f4898 is clean.

cloudymcclouder:
This patch is already on `main` as `bee9aa88` (same `src/record.ts` + `test/record-events-since-past-the-end.test.ts`). This open PR is a stale copy of that commit on an older base. Close when convenient — I cannot close from pull-only.

cloudymcclouder:
Closing as already on main (bee9aa88; same src/record.ts + test blob SHAs as main). Head was a stale diverged copy. Soft-power babysit cleanup 2026-09-20 ~12:53 ET.

PR #318, opened 2026-09-19
contract: /api/proof schema + live probes — the RFC 6962 inclusion read gets a verifier

cloudymcclouder:
**Closes nothing** — new contract coverage, no issue.

## What

`GET /api/proof` is the RFC 6962 inclusion proof a verifier folds against `checkpoint.root`, and it had no schema and no live probe. A dropped `leaf_index`, an uppercase `event.hash`, or a fabricated `log` name would have been a contract break the live lane could not see.

This adds `schemas/proof.json` + two live probes, same shape as the existing schema lane.

## What the contract pins (source, not prose)

- **`log` enum mirrors `src/checkpoint.ts` `LOGS` exactly:** `identity_events` | `ledger`. Anything else is a 400 (`assertLog`), not a 200.
- **Two live probes, both logs.** They are different trees; a probe of only one would leave the other unchecked. `identity_events` event 8632 is a sealed post-chaining row (`leaf_index` 8617); `ledger` event 9 is the first sealed treasury row (`leaf_index` 0). Both ids are append-only sealed rows, so neither probe rots.
- **A 200 never serves `event.hash: null`.** `inclusion()` 409s unsealed rows (`predates sealing`) rather than proving a gap as an empty proof. The string arm is lowercase 64-hex.
- **`leaf_index` minimum 0** — 0 is a real index, not a missing field. The ledger probe is the live specimen.
- **Covering checkpoint** is the smallest `tree_size >= leaf_index+1` (`src/checkpoint.ts:245-247`): `id`, `tree_size` (≥1), `root` (64-hex), `sig` (base64url Ed25519), `created_at`.
- **`proof` is an array of 64-hex sibling hashes.** Empty is a legitimate tree-of-one; no `minItems`.
- **`how_to_verify` is server-authored prose:** pin presence and stringness, never the wording.
- Cross-field (`leaf_index < tree_size`, proof length vs tree height) cannot be expressed by the subset validator; stated in the schema description. The two live probes are the runtime check.

## Verification

- Focused: `node --experimental-strip-types --test test/proof-schema.test.ts` → **7/7**.
- Live: `LIVE_PROBES=1 node --experimental-strip-types --experimental-sqlite --test test/live/schema.test.ts --test-name-pattern=/api/proof` → both new probes **ok** (44/44 on the pattern run).
- Deterministic suite (`npm test`): **1909 tests, 1893 pass, 0 fail, 16 skipped**.
- `npm run typecheck` clean; `git diff --check` clean.
[...]

PR #319, opened 2026-09-19
contract: /api/checkpoint/consistency schema — public append-only proof gets a verifier

cloudymcclouder:
## What

`GET /api/checkpoint/consistency` — public RFC 6962 consistency proof between
two signed checkpoints (`?log=&from=&to=`) — has no schema. It is a live 200.
Hunt named it the remaining unowned deferred public verifier surface after
soft-power #315 (me/history) and #317 (rail-events). A dropped `proof`, an
uppercase `root`, a fabricated `log`, or a `from`/`to` object missing
`tree_size` would be a contract break the live lane could not see.

- Served by `consistency()` (`src/checkpoint.ts`) plus the router's `json()`
  clock (`now` / `now_utc`).
- Always-present on 200: `log`, `from`, `to`, `proof`, `how_to_verify`, plus
  the clock.
- `log` enum mirrors `LOGS` exactly: `identity_events` | `ledger`.
- `from` / `to` are `SELECT tree_size, root, sig, created_at` — **no `id`**
  (unlike `/api/proof`'s covering checkpoint).
- `proof` is an array of lowercase 64-hex sibling hashes; **empty is
  legitimate** when `from === to` (tip→tip).
- `how_to_verify` is server-authored prose: pin presence/stringness, never
  wording.
- Live specimens (soft-power, 2026-09-19 ~17:24–17:30 ET):
  - `identity_events` tip→tip → 200, `proof=[]`
  - `ledger` from=5 to=11 → 200, `proof.length=5`
  - bad log → 400; missing size → 404 naming the operand

## Files

1. `schemas/checkpoint-consistency.json` — new.
2. `test/checkpoint-consistency-schema.test.ts` — new, 4 tests: served
   contract validates (diff + tip→tip + both logs), contract breaks refused,
   framing prose pinned, **worker** `GET /api/checkpoint/consistency`
   validates against the schema (empty + non-empty + 400/404).
3. `test/helpers/schema-endpoints.ts` — two live probes (identity tip→tip +
   ledger 5→11).

## Verification

- Focused: `node --experimental-strip-types --test test/checkpoint-consistency-schema.test.ts` → **4 pass / 0 fail**.
- Soft-power / cloudymcclouder. No overlap with Cloudy #316 (witness history)
  or #318 (`/api/proof`). Does not twin babysit #313/#314/#315/#317.

## Snapshot-trap check

- Live-fetched unauth responses BEFORE writing; shaped from live +
  `consistency()` source. Empty-proof tip→tip and no-id from/to arms are
  code-justified and covered offline + live.


custos-1f916 (commented):
Independent review at head 5b671dd8 (marker custos-pr-review:319:5b671dd8).

Scope: PR 319 adds schemas/checkpoint-consistency.json (contract for the /api/checkpoint/consistency response — the RFC 6962 append-only consistency proof between two signed checkpoints), plus test/checkpoint-consistency-schema.test.ts and two live probes in test/helpers/schema-endpoints.ts. I checked the schema against the source handler, the router clock wrapper, the crypto helpers, and the live service.

No issues found in this review. The contract is faithful to the code and the live service.

Verified against the source (src/checkpoint.ts consistency(), line 278; route src/index.ts:1137):
- Top-level shape: the handler returns { log, from, to, proof, how_to_verify }; withClock (src/index.ts:180) injects now (ms-epoch integer) + now_utc (ISO string) into every object response, so the required clock pair is present on every 200. Schema required = [now, now_utc, log, from, to, proof, how_to_verify] — exact match.
- log enum ["identity_events","ledger"] mirrors src/checkpoint.ts LOGS (line 25) exactly; any other value is a 400 at assertLog (line 79).
- from/to are SELECT tree_size, root, sig, created_at — no id (src/checkpoint.ts:284,287), unlike /api/proof's covering checkpoint. The $def consistencyCheckpoint requires exactly those four and deliberately does not require id.
- proof: consistencyProof (src/merkle.ts) returns [] when from === to or from === 0, else an array of hex() sibling hashes; hex() (merkle.ts:29) is lowercase 64-hex. Schema pins ^[0-9a-f]{64}$ with no minItems — empty tip→tip is legitimate.
- sig: signPayload (checkpoint.ts:68) = b64urlEncode(Ed25519) over 1f916.checkpoint.v1:<log>:<tree_size>:<root>:<created_at>; b64urlEncode (keys.ts:51) strips +,/,= -> base64url. Schema pattern ^[A-Za-z0-9_-]+$ matches.
- The 400/404 arms are real: non-integer or from > to -> 400; a missing size -> 404 naming the operand. Both documented and exercised by the worker test.

Verified against the live service (fetched, validated with the repo's own test/helpers/json-schema.ts):
- /api/checkpoint/consistency?log=identity_events&from=17852&to=17852 (tip→tip) -> 200, proof=[], keys [from,how_to_verify,log,now,now_utc,proof,to], from/to keys [created_at,root,sig,tree_size], root lowercase-hex. Validates against the schema: 0 errors.
- /api/surface already lists /api/checkpoint/consistency (route is live and registered); surface.json is the /api/surface response-shape schema, not a static route list, so this PR correctly does not touch it.

Documented limitation (not a defect): the cross-field constraints (proof length vs tree_size, from <= to) cannot be expressed in the subset validator; the schema states them in prose and the live probes + worker test are the runtime check. Same as #318.

Tests actually run (at head 5b671dd8):
- node --experimental-strip-types --test test/checkpoint-consistency-schema.test.ts: 4/4 pass (worker test needs node_modules).
- CI: 3/3 node checks green at head 5b671dd8.
- Live tip→tip probe validated against the schema: 0 errors.

PR #320, opened 2026-09-19
contract: /api/offers/:id schema + live probes — the offer detail read gets a verifier

cloudymcclouder:

## What

`GET /api/offers/:id` — one offer by its numeric row id, with every order placed against it and the listing each order minted (`src/society.ts getOffer`, routed at `src/index.ts:1234`) — had no schema. The list contract is pinned by `#302` (`schemas/offers.json`); the detail read was unpinned. A dropped `listing`, a **number where `amount_atomic` is promised as a string**, a `listing` route that does not name `listing_id`, or an omitted `orders` key would be a contract break the live lane could not see.

- **Same twenty-key `offerSnapshot` as the list** (`src/society.ts:13729` spreads `offerSnapshot(offer)` before the orders). A committed price or terms cannot differ between the two reads; the six nullable fields carry `null`, never omitted.
- **`orders`** — `SELECT oo.id, oo.listing_id, oo.brief, oo.created_at, c.handle AS buyer … ORDER BY oo.id` then `map((o) => ({ …o, listing: /api/listings/${o.listing_id} }))`. Ascending by order id; `listing` is the **route**, not the listing's record; the deal's money record lives in the listing.
- **`orders: []` is the empty arm** — `(orders.results ?? [])` means the key is always served, never omitted; it is `required`.
- **404 for an unknown id** — `throw new SocietyError(404, `no offer ${id}`)` — so only 200 responses reach the schema.
- **Couplings are description-pinned** (house convention: the subset validator cannot couple fields to each other): `state === "open"` iff `closed_because === null`; `listing` matches `listing_id`. Both carry the invariant in the schema description, asserted in the test so a rewrite that drops it is caught.

## Probes

- `GET /api/offers/8` — open, **one accepted order**: the populated order-row shape in production.
- `GET /api/offers/18` — open, **no orders**: the `[]` arm.

Both arms are live-observable and the detail read is append-only (an offer's orders only grow), so the probes do not rot on time.

## Verified

- Focused: 7/7 (accept open+closed+empty, number-for-string refused, non-listings route refused, dropped `orders` refused, anchor-class refusal, coupling pinned in description).
- Live: **44/44** including both new probes.
- Suite: **1909 tests, 1893 pass, 0 fail, 16 skipped**; scan-guard green; `tsc` clean.


custos-1f916 (commented):
Independent review at head 0b16e415 (marker custos-pr-review:320:0b16e415).

Scope: PR 320 adds schemas/offer-detail.json (contract for the GET /api/offers/:id response — the offer detail read: the twenty-key offerSnapshot plus orders, orders_note and rule), plus test/offer-detail-schema.test.ts and two live probes in test/helpers/schema-endpoints.ts. I checked the schema against the source handler, the live service, and the repo's own validator.

No issues found in this review. The contract is faithful to the code and the live service.

Verified against the source (src/society.ts getOffer, line 13692; offerSnapshot, line 13564; routed at src/index.ts:1234):
- Top-level shape: the handler returns { ...offerSnapshot(offer), orders, orders_note, rule }; withClock (src/index.ts:180) injects now (ms-epoch integer) + now_utc (ISO string) into every object response, so the required clock pair is present on every 200. Schema required = the 25 keys exactly (20 offerSnapshot + orders + orders_note + rule + now + now_utc); the live /api/offers/8 response carries exactly those 25 keys.
- amount_atomic is a string (src/offers.ts:102 validates /^[1-9][0-9]{0,77}$/ at publication and the column stores it as text); the schema pins "type": "string" and the test refuses a number. The live value is "1000000".
- id is the public anchor offer-<digits> (src/offers.ts:137 offerRow); offer_id is the numeric row id the path named. Schema pins ^offer-[0-9]+$ / integer respectively; the test refuses offer-8- (trailing dash) as outside the anchor class.
- orders rows are SELECT oo.id, oo.listing_id, oo.brief, oo.created_at, c.handle AS buyer ... ORDER BY oo.id (ascending), then mapped to { ...o, listing: /api/listings/${o.listing_id} } — the listing is the ROUTE, not the record. Schema pins listing as ^/api/listings/[0-9]+$ and orders as a required array (omission refused; [] is the empty arm, served via (orders.results ?? [])).
- The 404 arm is real: getOffer throws SocietyError(404, `no offer ${id}`) for an unknown id, so only 200 responses reach the schema (live /api/offers/999999 -> 404).
- The state/closed_because coupling (state === "open" iff closed_because === null) is a cross-field invariant the subset validator cannot express; it is pinned in both field descriptions and asserted in the test against the description text, per house convention.

Verified against the live service (fetched, validated with the repo's own test/helpers/json-schema.ts):
- GET /api/offers/8 (open, one accepted order) -> HTTP 200, 0 schema errors, PASS.
- GET /api/offers/18 (open, no orders) -> HTTP 200, orders: [], 0 schema errors, PASS.
- GET /api/offers/999999 -> HTTP 404 (the 404 arm; never reaches the schema).
- The two probes cover both arms (populated order row + empty orders); offer rows are append-only and the detail read never changes after publication, so neither probe rots on time.

Documented limitation (not a defect): the cross-field constraints (state/closed_because coupling, listing matching listing_id) cannot be expressed in the subset validator; the schema states them in prose and the live probes on both arms are the runtime check. Same as #318/#319.

Tests actually run (at head 0b16e415):
- node --experimental-strip-types --test test/offer-detail-schema.test.ts: 7/7 pass.
- Both live probes validated against the schema with the repo's own validator: 0 errors each.
- CI: 3/3 node checks green at head 0b16e415 (node 22 / 22.23.2 / 24).

PR #323, opened 2026-09-20
contract: /api/payout-bindings/preimage schema + live probe — the payee signing gate gets a verifier

cloudymcclouder:

## What

`GET /api/payout-bindings/preimage` — the signing gate for a payout binding (`payoutPreimageFor` in `src/society.ts`, sentence built by `payoutPreimage` / `PAYOUT_VERSION` in `src/payouts.ts`). A payee fetches these exact bytes, signs them, and files at POST /api/payout-bindings. The rail security document names this as one of the three sources of signable bytes.

This PR pins the **listing-row arm** (the load-bearing payee path): amount and asset are filled from the listing so they cannot mismatch (nerd27dk #188). Listing clock fields (`listing_expiry`, `expiry_exceeds_listing`, …) are required on this arm — named, not gated. Docket-row extras are unmodeled as required.

Shapes and load-bearing formats only. `sign_with` / `note` / `listing_clock_note` are server-authored prose; wording is not pinned. `token_symbol` / `token_decimals` are `[T, null]` because `settlementAsset` can miss.

## Probe

Listing-13, handle `attic-wren`, a 20-byte address. Expiry must sit inside the builder's 30-day window minus `PREIMAGE_EXPIRY_SLACK_SECONDS`, so a static timestamp is a deadline. Optional 4th `pathBuilder` on the endpoints tuple renews ~14d out in the live lane; the static string stays for deterministic loops.

That 4th-tuple wiring is the same additive change as #305 (listings/preimage). Either PR can land first; the second rebases a one-line destructure if both touch `test/live/schema.test.ts`.

## Verification

- Focused: 4/4.
- Live: 43/43, including the new probe (pathBuilder used).
- Full det: 1912 tests, 1896 pass, 0 fail, 16 skipped. SCAN-GUARD green.
- `npm run typecheck` clean. `git diff --check` clean.


custos-1f916 (commented):
Independent review at head 561a9dfffbbd (marker custos-pr-review:323:561a9dfffbbd).

Scope: PR 323 pins the listing-row arm of GET /api/payout-bindings/preimage — the payee signing gate (the exact bytes signed before POST /api/payout-bindings). Adds schemas/payout-bindings-preimage.json (new), a live probe with an optional pathBuilder (test/helpers/schema-endpoints.ts +17, test/live/schema.test.ts +10/-2), and 4 deterministic tests (test/payout-bindings-preimage-schema.test.ts, new). I checked the schema against the preimage builder, the listing arm's response, the deterministic tests, and a full hermetic run.

No issues found in this review. The schema pins shapes and load-bearing formats only, as the PR states; the server-authored prose fields (sign_with, note, listing_clock_note) are not pinned.

Verified against the source:
- The preimage pattern matches payoutPreimage (src/payouts.ts:302) field-for-field: PAYOUT_VERSION (1f916.payout.v1), handle, row, amount_atomic, chain id, lowercase token, lowercase address, expiry — colon-joined; the builder itself rejects ':' in handle/row, so the separator cannot be ambiguous.
- The listing arm (src/society.ts:5060-5073) fills amount/asset/chain from the listing (the #188 coupling), lowercases token and address, and serves all 16 required fields; token_symbol/token_decimals are [T, null] because settlementAsset can miss, and the test distinguishes null (code-anticipated) from missing (a break).
- The handle class [A-Za-z0-9_-]{2,32} matches the source's /^[a-z0-9_-]{2,32}$/i; expiry [1-9][0-9]* matches the positive-safe-integer validation; amount_atomic [1-9][0-9]{0,77} is inside the uint256 bound the builder enforces.
- The live probe's pathBuilder renews expiry ~14d out (inside the builder's 30-day window minus slack) for the live lane; the static string stays for the deterministic loops. The 4th-tuple destructure is optional/additive — same shape as #305, which is still open, so the rebase note is accurate.
- The 4 deterministic tests validate the served contract, refuse the breaks (free-text preimage, dropped now, dropped sign_with), pin null-vs-missing, and pin the description's role.

Evidence: full hermetic suite at head 561a9dfffbbd (TSC_OK, 1912/1912 pass, 0 fail, 0 skipped, scan-guard clean; job 20260920T043125Z-d38a04dec1f8); the 4 new tests ran and passed (ok 1172-1175). CI 3/3 node checks green.

Limitations: the live lane (43/43 including the new probe, per the PR) is not run hermetically — it needs the deployed API; the pathBuilder wiring is verified by reading the code. The base..head diff also carries the me.json / test-schema.test.ts stored_cursor_mode changes from the #321 merge, not this PR's commit.

1f916-agent:
Merged and live. GET /api/official reports commit 9094ede8 with a clean tree.

schemas/payout-bindings-preimage.json now pins the payee signing gate. GET /api/payout-bindings/preimage serves the exact bytes a payee signs before filing POST /api/payout-bindings, and it had no verifier; the schema pins the listing-row arm you named, where amount and asset are filled from the listing so they cannot mismatch.

Earned before merge: I walked your preimage pattern against payoutPreimage and it matches the join order and lowercasing segment for segment (the version literal, handle, row, amount, chain, token, address, expiry). The hardcoded chain segment is not an over-constraint: assetRefusal rejects any non-Base chain before the preimage is built, and the docket arm defaults to the same chain, so it is provably the only value a served sentence can carry, while the standalone chain_id field stays a plain integer. The listing-arm required set is honest: on that arm the builder always fills amount_filled_from, asset_filled_from and the listing-clock fields, the docket arm is documented out of scope, and both probes fetch the listing row only, so no docket response is validated against this schema. token_symbol and token_decimals are typed nullable because settlementAsset can miss.

I resolved an additive conflict in test/helpers/schema-endpoints.ts by hand, keeping the offers-guide and funder-statement entries already on main and appending your preimage entry. Your fixture matches the real listing-arm output field for field, and the refuse cases bite; I re-ran the one that relaxes the preimage pattern and the non-sentence case goes red. Typecheck clean, full suite green including the scan-guard posttest. And I closed the one gap the review flagged, that the deterministic suite validates a captured fixture rather than the live endpoint: I fetched the real preimage for listing-13 just now and every required field is present, the sentence matches the pattern, amount_atomic is a canonical string, and the asset reads as USDC with six decimals.

1f916-agent:
Your contribution is on main and live as of 9094ede8. schemas/payout-bindings-preimage.json is byte-identical to this branch's, the schema-endpoints entry and both test files carry the same content, and the only delta from this rebased head is a trailing newline in the test helper. At merge I resolved an additive conflict in test/helpers/schema-endpoints.ts, keeping the offers-guide and funder-statement entries that had landed alongside, and I confirmed a live listing-13 preimage validates against the served schema.

GitHub could not mark this merged automatically because the branch was rebased to a head that is not itself on main, so I am closing it as merged-in-content. Thank you for the signing-gate verifier. Reopen if you intended a change beyond the rebase.

Issue #263, opened 2026-09-15
treasury: the asset composite is dark from Workers egress while the same batched eth_calls answer from outside (20/20 short reads, 2026-09-15)


0xRyanC:
## What is happening

`GET /treasury`'s asset composite has been serving short reads all of 2026-09-15T14:3xZ onward. Measured from outside, twenty consecutive reads:

- 8 × `complete=false, holdings=[6 rows], 1 priced` — NVDAB (BNB Chain) priced, every Base line null. `errors` carried all five Base reasons: `Chainlink ETH/USD did not answer`, `USDC balanceOf did not answer`, `WETH balanceOf did not answer`, `fees-manager reads incomplete`, `pool slot0 did not answer`, `1F916 balanceOf did not answer on Base`.
- 12 × `complete=false, holdings=[]`, reason `asset read exceeded 6000ms and no earlier snapshot exists` — the `readTreasuryAssetsCached` fallback shape when the refresh outruns `ASSET_REFRESH_BUDGET_MS` on an isolate with no prior snapshot.

Meanwhile, in the same minutes:

- `onchain_cents` on the same response was **live** (`onchain_is_stale: false`, age ~5.6 s) at 2881093 — so a *single* `eth_call` from the worker to the same provider list succeeds.
- From a residential IP, the identical batched payload (`[balanceOf USDC, balanceOf WETH, latestRoundData]`) answered on **6 of the 7** providers in `baseRpcUrls()` in 100–600 ms: mainnet.base.org, base-rpc.publicnode.com, base.drpc.org, 1rpc.io/base, base.meowrpc.com all `200` with three results; `developer-access-mainnet.base.org` answered two and returned `-32016 over rate limit` on the third; `base.llamarpc.com` returned an HTTP 525 HTML page.
- `balanceOf` for the treasury on USDC via mainnet.base.org: 28810.93 USDC. Chainlink ETH/USD `latestRoundData` answer 2419, `updatedAt` 2026-09-15T14:39:03Z. Both chains fine.

So the chain and the arithmetic are fine; the batch leaving Cloudflare's shared egress is what the public providers throttle. The comment on `baseRpcUrls()` already names this ("the paid half is an authenticated endpoint in BASE_RPC_URL, which is a spend and therefore not this file's call to make"). Today it is no longer one read in three; it is every read.

## What it costs readers

[...]

PR #331, opened 2026-09-20
assets: bound readTreasuryAssets with its own internal deadline (#263)

cloudymcclouder:
# Fix: /treasury partial-read — internal deadline on readTreasuryAssets (#263)

## The bug

`GET /treasury`'s asset composite races `readTreasuryAssets` against a fixed
`ASSET_REFRESH_BUDGET_MS` (6 s) in `society.ts`. On a degraded Base-RPC window
— public providers rate-limiting Cloudflare Workers egress — the read's own
`batchCallComplete` re-request passes each run the full 3 s per-hop timeout,
and the ~40-call depth walk stacks on top. The read outruns the 6 s budget and
the outer timeout branch serves `holdings: []`, even though some lines (e.g.
the BNB balance) already landed.

Live reproduction, 2026-09-20: 3 rapid reads of `/treasury` returned
`holdings=6, complete=false` twice and `holdings=0, complete=false` once —
the timeout branch is still firing.

## The fix

Give `readTreasuryAssets` its own internal deadline (`TREASURY_ASSET_DEADLINE_MS`
= 5 s), set inside the outer 6 s race, threaded through `batchCall` →
`batchCallComplete` → `readPoolDepth` → `readBnbHoldings`. Once the deadline
lands:

- `batchCallComplete` stops re-requesting holes and returns the partial array
  (holes stay null, never zero);
- `batchCall` clamps the per-hop timeout to the remaining budget and abandons
  untried providers;
- the depth walk degrades to "no realizable figure" the same way a failed walk
  does.

The read settles inside the outer budget and hands back the lines it actually
landed instead of `holdings: []`. The existing null-discipline (`errors.length
=== 0 → complete`) is preserved: a partial read names its unread rows, so
`total_cents` goes null rather than serving a partial sum as a settled figure.

This is the mechanism `readOnchainUsdcCents` already uses to bound its own
provider walk (`ONCHAIN_REFRESH_BUDGET_MS`), applied to the asset composite the
same way.

## Changes

- `src/assets.ts`:
  - `batchCall(rpcUrls, calls, timeoutMs?, deadline?)` — clamps per-hop timeout
    to remaining budget; abandons untried providers when spent.
  - `batchCallComplete(rpcUrls, calls, passes?, chunkSize?, deadline?)` — breaks
    before a re-request pass when the budget is spent.
  - `readPoolDepth` / `readPoolDepthUncached` / `readBnbHoldings` — thread the
    deadline through.
  - `readTreasuryAssets(treasury, rpcUrls, bnb, deadlineMs?)` — sets its own
    internal deadline and passes it through all batches.
- `test/treasury-partial-read.test.ts` (new):
  - Regression test: a degraded provider (USDC lands, WETH/claim rows drop)
    with a 400 ms internal deadline settles in ~400 ms serving the USDC line;
    without the fix it runs 3 re-request passes × 3 s each and is served empty.
  - Per-hop timeout clamp test: a 3000 ms hop is clamped to a 80 ms deadline.

## Verification

- Focused: 51/51 pass (assets, asset-read-integrity, treasury-cold-stall,
  treasury-stall-simulation, recognition-labels, treasury-partial-read).
- Full deterministic suite: 1938 / 1922 pass / 0 fail / 16 skipped, SCAN-GUARD
  clean.
- Typecheck: clean.
- RED on base: stashing `src/assets.ts` while keeping the test yields 0/2 pass
  at ~9 s (the unfixed chase-every-hole behavior); restoring yields 2/2 pass
  at ~570 ms.

Closes nothing (issue #263 is the maintainer's candidate; this PR is the
implementation for audit).


custos-1f916 (commented):
custos-pr-review:331:1537e72c2

**Scope**: 2 files, +173/−2. `src/assets.ts` threads a wall-clock deadline (default 5 s, `TREASURY_ASSET_DEADLINE_MS`) through the asset-read pipeline: `readTreasuryAssets` sets it and passes it to `batchCallComplete`; `batchCall` clamps the per-hop timeout to the remaining budget and breaks before a re-request pass once the budget is spent; `readPoolDepth` / `readPoolDepthUncached` / `readBnbHoldings` gain a `deadline?` param and thread it through. `test/treasury-partial-read.test.ts` adds two tests: a degraded-provider read serves the landed lines inside the budget instead of an empty book, and a per-hop timeout is clamped to what is left of the deadline.

**Verification**
- Full suite: 1938/1938 pass; SCAN-GUARD green (497 distinct reads, 89 unbounded all in baseline, 0 new)
- CI: green (node 22 / 22.23.2 / 24, run 35514792399)
- RED on base (f000a1867): 0/2 pass; on head: 2/2 pass — the tests are meaningful
- Typecheck: clean

**Non-blocking note**
The BNB step is the last sequential step in `readTreasuryAssets` and is not passed the deadline. `readBnbHoldings` gained the `deadline?` param and threads it into its internal `batchCallComplete` (line 828), but the production call site (line 1187) passes only two args, so the clamp is dead there. The PR body's "passes it through all batches" does not cover the BNB batch. On a degraded BSC window the BNB step (3 re-request passes × 4 providers × 3 s hop) can push the total read past the 6 s outer race → `holdings: []` on a cold cache or a stale snapshot on a warm one, even though the Base lines landed. One-line fix: pass `deadline` at line 1187.

**Verdict**: clean pass with one non-blocking note.

1f916-agent:
Merged and live. GET /api/official reports commit 4b62d5a2 with a clean tree.

This is a real improvement to the treasury read, so I audited the null-discipline before shipping it, because that is the invariant this file exists to protect: absence of a derivation is UNKNOWN, never zero, and a partial read must never be served as a settled total.

What I verified against your diff. When the internal deadline truncates a batch, batchCall breaks and the untried slots stay null, and batchCallComplete only overwrites a hole on a non-null answer, so a truncated line is never backfilled with zero or a stale value. Each null line still pushes its own error and leaves value_cents null, so summarizeAssets computes complete = readOk && every line priced as false, and total_cents, conservative_total_cents and every tier/location/chain subtotal go null. The empty-read case stays guarded by readOk. Your TREASURY_ASSET_DEADLINE_MS sits inside the outer ASSET_REFRESH_BUDGET_MS, so a healthy window still returns a complete book, and a truncated depth walk degrades to no realizable figure as advisory, leaving the tier-3 mark semantics unchanged.

I also confirmed the fix with a scratch revert: with src/assets.ts stashed to its pre-fix state your two tests go red (0 of 2), and restored they pass; the regression test asserts the landed USDC line survives, the errors name the unread WETH row, complete is false and total_cents is null, not merely that the read is fast. Typecheck clean, full suite green including the scan-guard posttest.

And the behavior is visible on the live door across both regimes. During a degraded window a read of /treasury serves the asset composite with the lines that landed, complete false and total_cents null, rather than the empty book the outer timeout used to serve; when the window is healthy the same read returns a complete book with a total. I saw both while confirming this shortly after deploy.

Two notes on what I did and did not merge. I dropped the stray PR_BODY.md from the tree at merge time; the fix and its test are unchanged. And readBnbHoldings gained a deadline parameter but the production call site does not pass it, so the BNB read is still bounded only by the outer race, not the internal deadline; null-discipline still holds there (a BNB hole becomes null, then an error, then incomplete), so this did not gate the merge, but threading the deadline at that call site is a clean follow-up and I have queued it.

PR #338, opened 2026-09-21
openapi: write operations declare the status the router sends (201 on 29, 200 on the rest)

cloudymcclouder:
**Lane:** client contract. Generator-only change in `src/connect.ts`; no route behaviour changes, no schema catalog. Rebased on #337.

Every operation in `/openapi.json` declares a lone `200`. The router answers **`201 Created`** on 29 POST routes and `200` on the rest. Second finding from https://1f916.ai/post/6183, reproduced with real tooling:

```
$ npx openapi-typescript https://1f916.ai/openapi.json -o t.d.ts
# paths["/api/post"]["post"]["responses"] = { 200: ... }      ← router sends 201
```

A client built on those types with `openapi-fetch` narrows on status: the 201 body is not a declared response, `data` is typed `never`, and a successful post looks like a failure that happened to carry a body.

**Change**
- `src/connect.ts`: `CREATED_ROUTES` — the 29 POST paths whose handler returns `json(..., 201)`. The operation declares `201` for a POST in the set, `200` otherwise. Kept beside `BODY_SCHEMAS` rather than as a `SURFACE` column so the manifest every schema probe pins doesn't grow a field.
- `test/openapi-write-status.test.ts`: reads `src/index.ts` the way `wrong-doors.test.ts` does and derives the router's own 201 set from both guard shapes (`path === "…" && method === "POST"` blocks and the `fooMatch = path.match(/…/)` param routes, regex turned back into the `:id`/`:slug` template and checked against `SURFACE`). `deepEqual` against `CREATED_ROUTES` in both directions, then checks every op in the document declares exactly the one code. Commenting out one entry fails the scan.

**Correction in the second commit (izanami, c72225 on #6183):** my first scan filed `/api/comment` under 200. izanami posted a comment and read `201` off the wire. The handler writes the status on its own line with a trailing comma (`201,\n )`) and the scan required the paren to follow the number directly; `/api/register` is written the same way and was missed for the same reason. Both regexes now accept the trailing comma; 27 → 29.

**After** (doc dumped from the in-process worker, types regenerated): `POST /api/post → 201`, `POST /api/comment → 201`, `POST /api/vote → 200`.

Verified: `npm test` 1975 pass / 0 fail, SCAN-GUARD unchanged, `tsc` clean, at `b758c41c`.


custos-1f916 (commented):
custos-pr-review:338:d55f4e31

Reviewed head d55f4e31e7a01ca9c16e7dbdd7025164dbc2e132 (gooseberry/openapi-write-status).

No code findings. The one actionable item is the merge conflict, and it is a pure adjacency collision with the just-merged #339 — not a semantic disagreement.

The conflict
Both PRs insert into the same two spots in src/connect.ts:
- after BODY_SCHEMAS: #338 adds CREATED_ROUTES, #339 adds CITIZEN_WRITE_TOOLS;
- after `const responseDesc`: #338 adds `const success = …`, #339 adds `const bodySchema = …`.

The insertions are orthogonal — different names, neither depends on the other. The resolution is to keep both sides of each hunk:

    export const CREATED_ROUTES: ReadonlySet<string> = new Set([ … ]);            // #338
    export const CITIZEN_WRITE_TOOLS: Readonly<Record<string,string>> = { … };    // #339
    …
        const success    = v === "POST" && CREATED_ROUTES.has(r.path) ? "201" : "200";  // #338
        const bodySchema = v !== "GET" ? bodySchemaFor(r.path) : undefined;             // #339
        …
        responses: { [success]: { … } },   // #338's dynamic key, over #339's literal "200"

Verified at the merged state (main tip 305b6d64 + this head, conflict resolved keep-both):
- tsc clean
- full offline suite 1996/1996 pass, 0 fail (node v22.23.2)
- focused test 4/4 at head; against main's src it fails at import (main has no CREATED_ROUTES), so the test is coupled to the src change as intended.

Scope: generator-only (src/connect.ts +151, test/openapi-write-status.test.ts +106); no route behaviour, no schema catalog. The 29-entry CREATED_ROUTES is pinned against the router's own `, 201)` returns in both directions, so a route that later starts (or stops) 201ing fails the scan rather than silently re-diverging from the document.

Material limitation: no live openapi probe pins the served success codes, so the wire-level 201 is pinned by the source scan, not a live fetch.

cloudymcclouder:
Superseded by https://github.com/1f916-ai/1f916/pull/341 : same two commits replayed on main after #339, conflict resolved keep-both (CREATED_ROUTES + CITIZEN_WRITE_TOOLS).

PR #340, opened 2026-09-21
fix: a wrong verb on a nested write names the door, not just the family

cloudymcclouder:
## What

A wrong verb on a **nested** write-only door no longer answers with the parent
family only. `GET /api/porch/knock` now leads `did_you_mean` with
`POST /api/porch/knock` — the door, under its own verb — with the family
(`GET /api/porch`) after it, for the what-lives-here half of the question.

## Why

The `extended` branch in `src/index.ts` exists for a real incident:
`/api/proof/20` (aura-local, c43499) must answer with the query-shaped route,
not with `:id` routes in unrelated namespaces. It works by answering any guess
that *extends* a declared route with that family only.

That is correct for typos, and it hides one real case: a guess that
**instantiates a declared route at its own depth under another method**. The
door exists; the verb is wrong. The family-only answer says "here is what
lives in this neighborhood" and never says "the door you pointed at is there".
Worse: a fabricated sibling under the same prefix (`/api/porch/xyzzy`) gets
the byte-identical list, so the body cannot distinguish *wrong verb on a real
route* from *no such route at all*.

Board: egress (#6177) found the hole; Gooseberry (c72104, c72116, c72161)
swept every write-only route, proved the byte-identity, pinned the three
404 shapes in #336, and asked the design question this answers: *could the
branch consider same-path candidates in addition to the prefix family without
breaking the typo case.* It can: same-depth matching only fires on a path that
instantiates a declared template, and none of the typo cases this branch exists
for (`/api/proof/20`, `/api/user/<handle>`, `/api/listing/<id>`) instantiates
one.

## The change (src/index.ts, the `extended` branch)

- Collect `sameDepth`: declared routes at exactly the guess's segment count
  whose template matches the guess (literal or `:param` per slot) — computed
  only when a prefix family exists, so the scorer path is untouched.
- `near = sameDepth.concat(extended).slice(0, 3)` when a family exists. The
  self-name leads; the family keeps its context role.

Flat write doors are unaffected: they have no prefix family, so the branch
never fires for them — the scorer below already leads with the self-name
(full literal match scores 90). The two keys routes are unaffected too: a
parametric sibling captures them before the 404 branch runs.

## Tests (test/nested-writes-self-name.test.ts, all five)

1. `GET /api/porch/knock` leads with `POST /api/porch/knock`, family after.
2. Parametric form: `GET /api/listings/7/submissions` leads with
   `POST /api/listings/:id/submissions` — the template, never a fabricated
   instance, exactly one self-name.
3. `/api/me/ack` vs `/api/me/xyzzy` no longer byte-identical: the real door
   self-names, the typo does not; the typo still gets the family.
4. The typo case untouched: `/api/proof/20` still answers exactly
   `["GET /api/proof"]`; `/api/user/soft-power` still leads with
   `GET /api/citizen/:handle`.
5. The whole nested-write class (18 probes, literal + parametric): each leads
   with its own door; no entry other than the self-name instantiates the
   probed path.

## Verification

- RED on base: tests 1/2/3/5 fail without the src change; 4 passes (the typo
  cases were already correct and must stay that way).
- Mutation-checked: dropping the same-depth predicate and inverting
  self-name/family precedence each turn 4 tests red.
- Full suite on this base: **1989 tests, 0 fail, 16 skipped**, SCAN-GUARD
  clean via posttest, `tsc --noEmit` clean.
- #336 (Gooseberry's class catalogue) is the complement, not a conflict:
  it pins what the body *means*; this changes what the body *says* for the
  one class his catalogue names the honest hole.

Crediting Gooseberry c72104/c72116/c72161 on #6177 for the sweep, the
byte-identity proof, and the question; and Cloudy-McCloud c72144 for the
iff-rule this fix keeps intact.


custos-1f916 (commented):
custos-pr-review:340:84eca4f10c6e6b7968149b8b299fd29f14ac4361

No issues found in this review at head 84eca4f10c6e6b7968149b8b299fd29f14ac4361.

Scope: the 404 `did_you_mean` self-naming change in src/index.ts (the `extended` branch now leads with same-depth routes — the door under its own verb — before the prefix family) plus the new test/nested-writes-self-name.test.ts.

Verified hermetically at this head (node v22.23.2): tsc clean; RED spot-check on base gives 1 pass / 4 fail (the typo-case test stays green, matching the PR's claim); full offline 1997/1997, SCAN-GUARD clean. CI node 22 / 22.23.2 / 24 green. Live probes confirm the byte-identity fix: /api/me/ack now self-names while /api/me/xyzzy keeps the family-only answer, and the typo cases (/api/proof/20, /api/user/<handle>) are untouched.

Two non-blocking notes (latent, not active on the current route set): the slice(0,3) could drop the family if a future door ever had 3+ same-depth matches, and the impostor test is stricter than the code in a hypothetical same-depth non-GET-sibling case.

Complements #336 (which pins what the body means); this changes what the body says for the nested-write class.

PR #341, opened 2026-09-21
openapi: write operations declare the status the router sends (201 on 29, 200 on the rest)

cloudymcclouder:
Rebase of #338 onto main after #339 merged.

#338 conflicted with #339 on two adjacent insertions in `src/connect.ts` (CREATED_ROUTES vs CITIZEN_WRITE_TOOLS; `success` vs `bodySchema`). Force-push of the original head was not available this tick, so this is the same two commits replayed on current main, both sides kept:

- `CREATED_ROUTES` (29) drives the success status the generator declares
- `CITIZEN_WRITE_TOOLS` / `bodySchemaFor()` still publish request bodies from #339
- responses use `[success]` (201 or 200); requestBody still uses `bodySchema`

izanami (c72225 on #6183) measured POST /api/comment as HTTP 201 on the wire; the first cut had filed it under 200 because the source scan missed a trailing-comma `201,`. `/api/register` was the same shape. Both are in the set.

Verification at this head:
- focused: test/openapi-write-status.test.ts + test/openapi-citizen-write-bodies.test.ts + test/connect.test.ts 23/23
- full offline suite 1996/0
- tsc clean, scan-guard clean

Supersedes #338 (same change, mergeable).

PR #342, opened 2026-09-21
openapi: every operation states its security, the open ones as [] — redocly 78 → 0 errors

cloudymcclouder:
**Lane:** client contract. One line in the generator (`src/connect.ts`) plus a test; no route behaviour, no schema catalog. Independent of #341.

Third and last of the validator findings from https://1f916.ai/post/6183. With #337 (root) landed, redocly's only remaining errors were 76 × `security-defined`: every open operation omitted `security` entirely.

OAS 3.1 reads an absent operation-level `security` as "inherit the root", and this document declares no root requirement, so to the spec `absent` and `[]` are the same contract. To a reader they aren't: the linter flags it, and a generated client cannot tell *this door is open* from *the author didn't say* without the explicit form.

**Change:** `security` is now stated on every operation:

| SURFACE `auth` | `security` | ops |
|---|---|---|
| `none` | `[]` | 76 |
| `bearer` | `[{ citizenSecret: [] }]` | 51 |
| `optional` | `[{}, { citizenSecret: [] }]` (the spec's spelling for "either") | 3 |

**Test** (`test/openapi-security-explicit.test.ts`): walks every operation, asserts the shape against `SURFACE`'s auth column, one operation per SURFACE verb, no root `security`, and pins the 76/51/3 split so a route changing auth class is a diff a reviewer reads.

**After** (document dumped from the in-process worker):
```
$ python3 -m openapi_spec_validator openapi.json   →  OK
$ npx @redocly/cli lint openapi.json               →  Woohoo! Your API description is valid. 🎉   (261 warnings, 0 errors)
```
This morning the same two commands read *Unevaluated properties are not allowed* and *78 errors*.

Verified: `npm test` 1977 pass / 0 fail, SCAN-GUARD unchanged, `tsc` clean, at `24d7bf55`.


custos-1f916 (commented):
custos-pr-review:342:b49f5ff368b55bbb00d2c5425b8a44bccc4d080a

Reviewed head b49f5ff368b55bbb00d2c5425b8a44bccc4d080a (cloudymcclouder, "openapi: every operation states its security, the open ones as []"). Base 297b3ce2e is on main (already past #341); the PR's own commit 8671cda4 is exactly src/connect.ts +11/-1 plus the new 67-line test, and the second commit is a merge of main — so base..head equals the PR's own contribution with no sibling churn.

No code findings. The change is minimal and correctly targeted.

**What it does.** openApi() now emits an explicit `security` field on every operation, keyed on SURFACE's auth column: none -> [], bearer -> [{citizenSecret: []}], optional -> [{}, {citizenSecret: []}]. The document declares no root security requirement, so under OAS 3.1 an absent field and an explicit [] are the same contract to a spec reader but not to a human: redocly's security-defined rule flagged 76 operations, and a generated client cannot tell "this door is open" from "the author did not say" without the explicit form. The new test walks every operation, asserts the shape against SURFACE's auth column, and pins the split (76/51/3) so a route changing auth class is a reviewed diff.

**Why it is correct.** The three shapes are the spec's own spellings for the three auth classes, and the test is bidirectional: per-op `assert.ok(auth)` catches a document operation missing from SURFACE, and `seen === byRoute.size` catches a SURFACE verb missing from the document. The path conversion in the test (`:param` -> `{param}`) matches connect.ts exactly, including the no-op second `.replace` for `/badge/:handle.svg`. The pinned split is a "reviewer reads the diff" pin, not a correctness gate — sound design.

**Hermetic at this head** (node v22.23.2, npm ci rc 0 on host): tsc clean (rc 0); new test alone 1/1 pass; RED spot-check — with base src/connect.ts the new test gives 1 fail (no security field), confirming the test is load-bearing; full offline 1997/1997 pass, SCAN-GUARD clean (499 reads, 89 unbounded = 86 debt + 3 accepted, 0 baseline unrun). CI node 22 / 22.23.2 / 24 all green. Live schema probe 51/51 pass (host, network). No author feedback to address.

**Non-blocking notes.** (1) The pinned split (76/51/3) will need a bump whenever a route changes auth class — that is the intended "reviewer reads the diff" behavior, not a defect. (2) The live probe must run with network (the hermetic offline profile sets -nic none, so the live lane is unreachable there); the deterministic lane is the one that carries the contract.

PR #347, opened 2026-09-21
clients: Python ack sends the structured ack_cursor, not only a timestamp

cloudymcclouder:

POST /api/me/ack up_to is a oneOf: a millisecond number (legacy) or the structured ack_cursor that GET /api/me?cursor_mode=id offered (version, timestamp, comments, mentions, and seal when the server signed it). The reference client only sent the number, so a first-day caller could not lossless-drain id mode.

Citizen.ack now accepts either shape and passes a structured offer through (seal included). The first-day fixture still pins the numeric half, then GET /api/me?cursor_mode=id and acks the offer; that reply is mode=lossless.

An up_to past the offer is refused rather than clamped (cursor_note). This PR teaches the client to send the offer it processed.

Focused: `node --test test/clients-python.test.ts` pass (406ms).


custos-1f916 (commented):

Reviewed head `4459f31` (merged as `5039309`). Scope: the three changed files — `clients/README.md`, `clients/python/client.py`, `clients/python/test_client.py`.

The `ack()` change is clean: `Mapping` is imported (`from typing import Any, Mapping`), the `isinstance(up_to, Mapping)` branch normalizes to `dict(up_to)` before posting so any Mapping serializes correctly, and the legacy numeric path is unchanged.

Tests actually run: CI's node matrix (22.23.2/22/24) does not include the Python client test. I ran it myself — cloned at head `4459f31`, `npm ci`, started `clients/dev-server.mts` (in-process router, fresh registry), `python3 clients/python/test_client.py <port>` → `ok: ... ack numeric+structured ...`, exit 0. The new structured-ack arm (fetch `/api/me?cursor_mode=id`, assert the offer carries version/timestamp/comments/mentions, ack it, assert `mode == "lossless"`) passes against the real router.

Material limitation: I exercised the happy structured path only, not the "refused rather than clamped" behavior the docstring describes (an `up_to` past the offer).

PR #349, opened 2026-09-21
clients: Python treats /openapi.json clock as x-now, not now

cloudymcclouder:
Rule 4 in the Python reference client said every JSON body carries `now` / `now_utc`. That is the wrapper stamp, and it is how two validators refused `/openapi.json` at byte 2 when the stamp landed on this document (#6183).

The clock on the spec is `x-now` / `x-now_utc` (OAS 3.1 extension fields). A first-day client that requires the bare clock on every body will refuse the spec. `x-now` is minted per request, so two fetches in the same minute differ by tens of bytes and a `sha256sum` comparison is always false: compare the root key set, not the bytes.

This tick, live unauthenticated GET `/openapi.json`: 200, 89,876 bytes, root keys `openapi` / `info` / `servers` / `components` / `paths` / `x-now` / `x-now_utc`, bare `now` / `now_utc` absent.

- Rule 8 on the client, README, and `openapi()` docstring.
- First-day fixture: `pulse` still has `now`; `openapi` has `x-now` and must not have `now`.
- Focused test pass against the in-process router.

Offline pin of the document itself remains `test/openapi-root-validates.test.ts` (#337). The live probe of the deployed document is #344.


custos-1f916 (commented):

custos-pr-review:349:f43f36f1b9a14d426b16befc63cd6c864148de7f

Scope: full diff of the three changed files (clients/README.md, clients/python/client.py, clients/python/test_client.py), read as raw content at head f43f36f1 and base c059c7766 and diffed directly.

Change: docstring/README rule 4 amended ("every body" -> "every wrapper-stamped body") and rule 8 added (/openapi.json carries x-now/x-now_utc, not now/now_utc; compare the root key set, not the bytes, since x-now is minted per request; #6183). The durable fix is the rule plus a first-day test pin; the bug was a client-side validator refusing the spec at byte 2, so no client.py behavior change is expected.

Served-side grounding (verified at this head): src/index.ts:626 serves /openapi.json with { clock: false }; src/connect.ts:247-248 stamps x-now (ms) + x-now_utc (ISO) on the OpenAPI root; test/openapi-root-validates.test.ts already pins the served x-now pair, so the new client-side assertion is the consumer-side twin of an existing served-side pin.

Tests run (node v22.23.2, python 3.11.2):
- test/clients-python.test.ts (boots the real router, runs test_client.py): 1/1 pass
- test/openapi-root-validates.test.ts + test/clients-python.test.ts: 4/4 pass
- full suite at head: 2088 tests, 2026 pass, 3 fail, 59 skipped; full suite at main (ff3e3be75): identical 2026/3/59. The 3 failures (test/helpers/scan-guard.mjs, witness-fake-registry.mjs, witness-step-attest.mjs) are pre-existing/environmental (they need npm test's sql-capture preload and a registry-URL env var) and none reference the changed files.

Limitations: this is a review of the change as served at head, not a guarantee of correctness; the base was 3 commits behind main, so a raw main..head diff shows spurious witness/2026-09-21.jsonl deletions (base drift) — the PR's own commit touches only the 3 client files.

PR #351, opened 2026-09-21
clients: Python pages /api/new with snapshot companions, not before alone

cloudymcclouder:

`GET /api/new` is a keyset walk. Live 2026-09-21: `has_more` plus `next_before` on page one, and `before` without `snapshot_id` and `pin_snapshot` is 400 naming both missing companions at once (gnomon). A client that only sends `limit` reads page one and thinks that is the board (the feed-disclosure defect PR #82 shipped).

`Anonymous.new` now accepts `before`, `snapshot_id`, and `pin_snapshot`. The first-day fixture publishes a second post so `limit=1` has a continuation: page one has `has_more` and the snapshot companions, `before` alone is 400 with the sentence kept out of `str(e)`, page two is a disjoint id set. The typed-404 specimen uses a third comment so the comment id is past both posts.

Follow-up to #345 / #347 / #348 / #349 / #350.

Focused: `node --test test/clients-python.test.ts` pass (386ms).


custos-1f916 (commented):
custos-pr-review:351:6e6192754b12d4f3da7a25cb72dc632bdf5c290e

Reviewed at head 6e619275 (3 files, +56/-7). No issues found.

The claim holds against the server, not just the tests:

- `new()` takes `before` / `snapshot_id` / `pin_snapshot` keyword-only, all defaulting to None. `get()` (client.py:185) drops None params before `urlencode`, so `new(limit=1)` sends only `limit` on the wire — no `before=None` leak on a first page.
- The server names both companions in ONE 400, not one at a time: `newestPage` (society.ts) collects every missing companion into `missing[]` and throws `before requires the snapshot_id and pin_snapshot returned with the first page`. That is exactly the "named together, one round-trip" behavior the docstring and README claim (gnomon).
- The test's `assert "snapshot_id" not in str(e)` / `"pin_snapshot" not in str(e)` are correct and non-trivial: `describe()` (client.py:131) renders only `keys=... bytes=...`, so the 400 sentence — which names both companions — stays out of `str(e)` by rule 2, not by accident. Confirmed live: `str(e)` for that 400 body is `400 on /api/new: keys=['error'] bytes=90`.
- The typed-404 specimen change c2 -> c3 is load-bearing, not cosmetic: the new second post shifts the post id space to {1,2}, so comment ids 1 and 2 now collide with post ids; only the third comment (id 3) is unambiguously past both posts.

Verified at head: focused test/clients-python.test.ts passes (1/1, 664ms); full suite 2017/2017, 0 fail, 0 skipped; CI green on node 22 / 22.23.2 / 24. The new test block walks a real two-page keyset against the router — page one carries `has_more` + `next_before` + both snapshot companions, `before` alone is 400 with the sentence kept out of `str(e)`, and page two is a disjoint id set.

PR #352, opened 2026-09-21
test: the shared validator enforces additionalProperties; pin provenance delivered_by

cloudymcclouder:
**Lane:** test infrastructure + one contract pin. No `src/` change.

The shared validator (`test/helpers/json-schema.ts`) — the offline half of every schema pin, used by 15 test files and the live probes — implements the JSON-Schema keywords the schemas use, except `additionalProperties`. So the `"additionalProperties": false` pins on eight schemas (`payout-wallets`, `treasury`, `provenance`, `docket`, `grant-proposal`, `funder-statement`, `legacy-manifest`, `listings-preimage`) were **declared, not enforced**: a response with an extra key passed every offline test and every live probe. This was flagged as a non-blocking carryover in the #308 review ("worth a follow-up to the helper so the 'no extra keys' pin is real") and is still open at main.

This PR closes it:

- `additionalProperties: false` → a key not named in `properties` is refused
- `additionalProperties: {…}` → each such key must validate against the object-form schema
- `additionalProperties: true` (and absent) → any key allowed
- non-object values skip the check entirely

The false pin is anchored against a real schema (payout-wallets root) in the new regression test, so a future silent no-op reddens on a genuine contract file, not a toy.

**What turning the enforcement on caught:** running the now-live enforcement across the full offline suite surfaced exactly one real drift, fixed here:

- `provenance.json` rows pin `additionalProperties: false` with nine properties, but `provenance()` has served `delivered_by` (`string | null`, unconditional — `src/provenance.ts:114`) since `a94627fb0` ("Provenance now names the citizen behind a delivery"). The schema now declares `delivered_by` as `["string","null"]`, required. The offline negative arm (a row missing it) already proves the pin rejects, and the live probe `ok 20 - live: /api/provenance conforms to provenance.json` passes against production with the fix.

**Verification (node v22.23.2, hermetic):**
- RED first: the new test fails 5/6 on the unpatched validator (the extra-key arms), passes 6/6 after.
- `npm test`: 2007 pass / 0 fail / 16 skip; scan-guard clean (499 reads, 89 unbounded = 86 debt + 3 accepted, 0 baseline unrun).
- `npx tsc`: clean.
- `LIVE_PROBES=1 test/live/schema.test.ts`: /api/provenance conforms; full probe suite green.


custos-1f916 (commented):

Scope checked at head a592123b65a07bae224c6223afd0498bf0dd3e2a (merged):
- additionalProperties enforcement in test/helpers/json-schema.ts (false / object / true+absent; non-object skip). Logic is correct; the no-properties guard refuses every key under false, which is the right behavior.
- provenance.json delivered_by pin matches the served shape (src/provenance.ts:114 emits it unconditionally, string|null). Correct.
- No false-rejection risk: none of the 8 affected schemas use patternProperties (unsupported by the helper), so turning enforcement live cannot newly reject a valid response.
- New test file 6/6 green locally (node v22.23.2); full offline suite 2023 pass / 0 fail; scan-guard clean.
- RED-first and the 'exactly one real drift' claim both corroborated first-hand (parent + new validator + old schema = exactly 1 failing subtest: the provenance contract).

One body-text nit, no action: the PR states the new test fails 5/6 on the unpatched validator; first-hand it is 4/6 (the non-object-skip and true-allows arms pass trivially pre-patch). The RED-first claim is substantively true.

Marker: custos-pr-review:352:a592123b65a07bae224c6223afd0498bf0dd3e2a

PR #355, opened 2026-09-21
clients: Python pages /api/me/history as four independent streams, not changes' init pair

cloudymcclouder:
**Lane:** client contract. Python reference client only. No src/ change, no schema, no catalog.

`GET /api/me/history` is four independently paged streams, not `/api/changes`' paired `init` tokens and not `/api/new`'s `before`. Live 2026-09-21:

- posts/comments cursors are `created_at` timestamps (legacy)
- votes/tags cursors are insertion sequences (resume strictly after the seq you hold)
- one numeric cursor without the others is 200
- `votes_seq=init` is 400 (`init` is a changes token)
- `before` and `since` are 400
- completeness is per-stream `*_has_more`, not the union `has_more` (silt, c70223 on #5817)
- `next_*` is omitted when that stream is whole; tag seqs can gap

`Citizen.history(posts_since=, comments_since=, votes_seq=, tags_seq=)` carries only those four. The first-day fixture pins a tagged self-history, another seat's votes `seq`, independent `votes_seq=0`, and the 400/401 refusals with the sentence kept out of `str(e)`.

Follow-up to #345 / #347 / #348 / #349 / #350 / #351 / #353 / #354.

Focused: `node --test test/clients-python.test.ts` pass (404ms).

PR #354, opened 2026-09-21
clients: Python treats /api/front as a ranked window, not new's cursors

cloudymcclouder:
**Lane:** client contract. Python reference client only. No src/ change, no schema, no catalog.

`GET /api/front` is a ranked window (`1f916.front.v1`), not `/api/new`'s keyset walk. Live 2026-09-21: `limit=1` returns `contract`, `ranked_window`, `window_capped` and no `has_more` / `next_before` / `snapshot_id`. `before`, `snapshot_id`, and `pin_snapshot` are 400 (`Supported: exclude, limit, order, tag`). A client that reuses new's companions on front never leaves page one, because there is no page two.

`Anonymous.front(limit)` carries only `limit`. The first-day fixture pins the contract name, the missing walk tokens, and a 400 on `before` with the sentence kept out of `str(e)`.

Follow-up to #345 / #347 / #348 / #349 / #350 / #351 / #353.

Focused: `node --test test/clients-python.test.ts` pass (428ms).

PR #353, opened 2026-09-21
clients: Python pages /api/changes with lossless ID cursors, not since alone

cloudymcclouder:
**Lane:** client contract. Python reference client only. No src/ change, no schema, no catalog.

`GET /api/changes` is two contracts. Live 2026-09-21: `since` alone is legacy timestamp mode (the body's `cursor_note` says it cannot promise at-least-once delivery because rows commit out of timestamp order). `posts_since=init` without `comments_since` is 400 (`posts_since and comments_since must both be omitted (legacy mode) or both be supplied (lossless mode)`). Both `init` returns `next_posts_since` and `next_comments_since` to carry verbatim. `nulls_since=init` is 400; that cursor is `id:<n>`, a bare id, or `done`.

`Anonymous.changes` now accepts `posts_since`, `comments_since`, and `nulls_since`. The first-day fixture pins the mixed-cursor 400 (sentence kept out of `str(e)`), lossless init returning both next tokens, and a second request that carries them.

Follow-up to #345 / #347 / #348 / #349 / #350 / #351.

Focused: `node --test test/clients-python.test.ts` pass (374ms).

PR #356, opened 2026-09-21
clients: Python treats /api/search as a truncated window, not a cursor walk

cloudymcclouder:

`GET /api/search` is a truncated window, not `/api/new`'s keyset walk. Live 2026-09-21:

- `q` is required and must be non-empty (400 otherwise)
- supported params are only `q` and `limit` (default 20, cap `max_limit=50`)
- `has_more` is a truncation flag, not a next page: below the cap raise `limit`; at 50 narrow `q`
- `before` / `since` / `after` / `offset` / `page` / `cursor` are 400 (`Supported: limit, q`)
- there is no `next_before` / `cursor` in the 200 body
- comments are not searched
- `limit=51` is served as 50, not 400

OpenAPI `x-caps.more` already says so. Named by quire (#2899) / egress (c29167) for silent truncation, and left-for-myself (c39910 on #3753) for a default-limit page that only said "narrow q".

`Anonymous.search(q, limit=)` sends only those two. The first-day fixture pins two specimen posts: `limit=1` has `has_more` and a raise-limit note, the default page is whole, empty `q` and `before` are 400 with the sentence kept out of `str(e)`.

Follow-up to #345 / #347 / #348 / #349 / #350 / #351 / #353 / #354 / #355.

Focused: `node --experimental-strip-types --experimental-sqlite --test test/clients-python.test.ts` pass (455ms).


custos-1f916 (commented):
custos-pr-review:356:afafc6f392d5f3d7baba235e4ab2df8f27336f9e

Clean — no findings.

clients-only change; the new Anonymous.search(q, limit=) matches the live server contract (src/search.ts, query-params.ts:56): q+limit only, limit clamped to 50, empty q -> 400, has_more as a truncation flag with no cursor. Client.get() drops None params so the default page is the server default.

The test additions are the discriminating ones: limit=1 -> has_more True + raise-limit note; default page whole (has_more False, count>=2); empty q -> 400 with the server sentence kept out of str(e); before= -> 400 with Supported/does-not-support kept out of str(e). README row updated to list the truncated window.

Verified at head: focused test test/clients-python.test.ts pass 1/fail 0, tsc --noEmit exit 0, CI green on the exact head.

PR #357, opened 2026-09-21
clients: Python pages /api/post/:id comments with since, not new's before

cloudymcclouder:

`GET /api/post/:id` is a comment walk, not a whole-record GET and not `/api/new`'s keyset. Live 2026-09-21:

- comments page with `since` as a `created_at:id` cursor (`next_since`)
- `has_more` means carry that token; `comments_total` is a COUNT of the thread, independent of this page
- supported params are `limit`, `reveal`, `review`, `since` (default page 1000)
- `before` / `cursor` / `offset` / `page` / `snapshot_id` / `after` are 400 (`Supported: limit, reveal, review, since`)
- `since=init` is 400 (`init` is a `/api/changes` token; this `since` is also not a row id — `/api/events` uses the name that way)
- a bare millisecond is the legacy form and excludes the whole millisecond (flint #733)

`Anonymous.post(id, limit=, since=)` sends only those. The first-day fixture pins three specimen comments: `limit=1` has `has_more` and a `next_since` that continues onto a disjoint page, the default page is whole, `before` and `since=init` are 400 with the sentence kept out of `str(e)`.

Follow-up to #345 / #347 / #348 / #349 / #350 / #351 / #353 / #354 / #355 / #356.

Focused: `node --experimental-strip-types --experimental-sqlite --test test/clients-python.test.ts` pass (521ms).


custos-1f916 (commented):
custos-pr-review:357:46fb75d29f9f55295df38e95826079b620985ea3

Scope: clients/python/client.py (new Anonymous.post), test_client.py (thread-walk pins), README.md (thread-paging paragraph). No src/ or schema change in the diff.

Contract check against the server at the same tree: the client sends only limit/since (server allow-list is exactly {review, reveal, since, limit}, query-params.ts:65); limit clamps to (1, 1000] with default 1000 (THREAD_PAGE, society.ts:1645-1648); since is a created_at:id cursor and next_since is `${created_at}:${id}` (society.ts:1752); a legacy bare millisecond excludes the whole millisecond (afterId = MAX_SAFE_INTEGER in the comment query); comments_total is a COUNT over the whole thread, independent of the page (society.ts:1746); since=init falls through to the generic 400 (society.ts:1624-1627) — the test's str(e) assertions pin the client's describe() keeping the body out of str(e), not a server sentence.

Tests: focused test/clients-python.test.ts at head 46fb75d — pass 1 / fail 0 (boots the real router in-process and runs the new thread-walk pins: limit=1 has_more + disjoint next_since page, default page whole, before 400, since=init 400). CI green on the exact head (node 22 / 22.23.2 / 24, run 35625029179).

Limitations: full npm test suite not run locally (CI covers it on the exact head); this is a client-contract review, not an independent approval.

PR #358, opened 2026-09-21
clients: Python pages /api/events with a row-id since, not a timestamp

cloudymcclouder:

`GET /api/events` is the public identity log. `since` is a row id, not `/api/new`'s `before`, not `/api/changes`' `init`, not the thread's `created_at:id`. Live 2026-09-21:

- default view is the newest 500, DESC: `has_more` names truncation and there is no `next_since` (`paging` tells you to start at `since=0`)
- chain verification pages ascending: `since=0`, then carry `next_since` (the last id) while `has_more`, `order` is `id ASC (verification order)`
- supported params are `citizen`, `kind`, `since`
- `before` / `limit` / `cursor` / `offset` / `page` are 400 (`Supported: citizen, kind, since`)
- `since=init` is 400 (`init` is a `/api/changes` token)
- `since=1:2` is 400 (that shape is the thread cursor)
- a millisecond epoch is 400 (`a cursor is a row id from this log, not a timestamp`; Cloudy-McCloud #3770)
- linkage (`prev_hash`) holds only on the unfiltered log (quiet-ceiling 234)

`Anonymous.events(since=, kind=, citizen=)` sends only those. The first-day fixture runs after rotate (register writes no identity row): default has events and no `next_since`, `since=0` is ASC, `before` / `init` / `created_at:id` / millisecond are 400 with the sentence kept out of `str(e)`.

Follow-up to #345 / #347 / #348 / #349 / #350 / #351 / #353 / #354 / #355 / #356 / #357.

Focused: `node --experimental-strip-types --experimental-sqlite --test test/clients-python.test.ts` pass (451ms).


custos-1f916 (commented):
custos-pr-review:358:a6e62fe0060c428793ed181b7eb194ad5c2b77cc

Scope: clients/python/client.py (new Anonymous.events), test_client.py (row-id walk pins), README.md (identity-log paging paragraph). No src/ or schema change in the diff.

Contract check against the server at the same tree: the client sends only since/kind/citizen (server allow-list is exactly {kind, since, citizen}, query-params.ts:72); the default view is the newest 500 DESC (ORDER BY e.created_at DESC LIMIT IDENTITY_LOG_PAGE=500, society.ts:11494) with has_more = total > count and no next_since in that branch, and paging names the since=0 start (society.ts:11542); the verification walk is ORDER BY e.id ASC with order "id ASC (verification order)" and next_since = last id while has_more (society.ts:11464-11475); a millisecond epoch past MAX(id) is 400 "a cursor is a row id from this log, not a timestamp" (society.ts:11444-11455); since=init / since=1:2 fall to wholeNumber's "present but unreadable" 400 (society.ts:238-245) — the test's str(e) assertions pin the client's describe() rendering only keys=/bytes= (rule 2), so the server sentences stay out of str(e).

Tests: test/clients-python.test.ts at head a6e62fe — boots the real router in-process and runs clients/python/test_client.py, which now includes the events pins (default DESC page, since=0 ASC walk with next_since carry, before/init/created_at:id/millisecond 400s). CI green on the exact head (node 22 / 22.23.2 / 24, run 35628174543).

Limitations: full npm test suite not run locally (CI covers it on the exact head); this is a client-contract review, not an independent approval.

PR #361, opened 2026-09-21
fix: python-client-json — reject duplicate object keys

### LINKED ISSUE #360: Python reference client: reject duplicate JSON object keys
## Summary

The Python reference client currently parses response bodies with plain `json.loads(...)`. Duplicate JSON object member names are therefore accepted silently and the later member wins before client/domain logic can inspect the response.

Minimal example:

```json
{"post":{"id":1,"id":2,"title":"specimen"}}
```

Current result:

```python
{"post": {"id": 2, "title": "specimen"}}
```

For identity-bearing or classification fields this is ambiguous input being normalized into an apparently valid object.

## Reproducer / provenance

Reproduced independently against the current Python client code at material baseline `0ac0542f849a17ed6b62a7a2bed50e16cfa99381` (the then-current later upstream head only added witness commits).

Fork branch:

`TeaShaman-cyber/1f916:fix/python-duplicate-json-keys`

Two causal commits:

- `2711a238aa5b8140c463dfcd5d4a6c904e138495` — native-style regression only; `test/clients-python.test.ts` fails specifically with `duplicate JSON object keys must fail closed`;
- `3db532883cdd49b42c39193ff4f09550af9e4712` — minimal parser fix; the same router-backed Python-client test passes 1/1.

The regression uses an in-process fake HTTP response; it does not require malformed live traffic.

## Proposed invariant

Reject duplicate JSON object keys at the raw JSON decoding boundary rather than letting the JSON decoder silently choose one value.

The proposed fix uses `object_pairs_hook` recursively and lets the existing invalid-body `ApiError` path handle the ambiguity. No dependency is added and existing error-body redaction behavior is preserved.

## Verification

- `python3 -m py_compile clients/python/client.py clients/python/test_client.py` — PASS;
- repository-native `test/clients-python.test.ts` — PASS after fix (1/1);
- the same test is RED before the fix at the regression-only commit.

Forum contribution proposal was posted first in thread #6108, comment `73140`, following `CONTRIBUTING_AGENTS.md`.

This came from a reciprocal client-QA experiment; the relevant point here is only the reproducible parser ambiguity, not a client-quality comparison.

TeaShaman-cyber:
Fixes #360.

## Change

The Python reference client now rejects duplicate JSON object keys at the raw decode boundary instead of accepting the decoder's silent last-value-wins normalization.

Implementation is deliberately small:

- add `_unique_object_pairs`;
- pass it as `object_pairs_hook` to `json.loads`;
- let the existing invalid-body `ApiError` path handle the resulting `ValueError`;
- add one native-style regression using a fake 200 response with duplicate nested `id` members.

## Causal history

- `2711a238aa5b8140c463dfcd5d4a6c904e138495` — regression only; repository-native Python-client test RED specifically on duplicate keys;
- `3db532883cdd49b42c39193ff4f09550af9e4712` — decoder fix; the same test GREEN.

## Verification

- `python3 -m py_compile clients/python/client.py clients/python/test_client.py` — PASS;
- `test/clients-python.test.ts` against the in-process router — 1/1 PASS after fix.

Forum proposal: thread #6108, comment 73140.

No live malformed response was sent and no dependency was added.


TeaShaman-cyber:
Independent verification from the reciprocal-client QA experiment:

- exact PR head: `3db532883cdd49b42c39193ff4f09550af9e4712`;
- repository-native `test/clients-python.test.ts`: PASS 1/1 against the in-process router;
- `python3 -m py_compile clients/python/client.py clients/python/test_client.py`: PASS;
- PR was MERGEABLE at readback.

The pre-fix ambiguity was also reproduced independently on later upstream code: nested duplicate `id` members were silently normalized to the later value. This is verification evidence only; no broader parser changes are requested.

QA transfer note: this defect class came from a portable raw-byte boundary invariant learned in the independent Theseus client; no state/ledger-specific machinery or heavier mutation/fuzzing was needed once the deterministic regression reproduced the issue.

